PR #8, opened 2025-02-26
[WIP] topk-js v1

jerguslejko:


topk-js/.github/workflows/CI.yml
```diff
@@ -0,0 +1,484 @@
+name: CI
```
jerguslejko, line 1: nit: this file should be moved to root `.github` folder

topk-js/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+edition = "2021"
+name = "topk-js"
+description = "TopK JavaScript SDK with TypeScript support"
+version = "0.0.0"
```
jerguslejko, line 5: note: this will need to follow the same versioning scheme as `topk-py` (and `topk-rs`) since they all share the same code

topk-js/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+repository = "https://github.com/fafolabs/topk-sdk"
+homepage = "https://topk.io"
+documentation = "https://docs.topk.io"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+# Default enable napi4 feature, see https://nodejs.org/api/n-api.html#node-api-version-matrix
+napi = { version = "3.0.0-alpha.33", default-features = false, features = ["napi4", "tokio_rt", "serde-json"] }
```
jerguslejko, line 15: nit: let's add a TODO to upgrade to a stable version when it's released. @petertoth any idea when that's going to happen?

topk-js/index.d.ts
```diff
@@ -0,0 +1,192 @@
...
+  delete(ids: Array<string>): Promise<number>
+}
+
+export declare class CollectionsClient {
+  list(): Promise<Array<Collection>>
+  create(options: CreateCollectionOptions): Promise<Collection>
+  delete(name: string): Promise<void>
+}
+
+export declare class Expr {
```
jerguslejko, line 21: the exprs need to be split into:
    * `SelectExpr` which contains either `LogicalExpr` or `FunctionExpr`. only accepted in `select()`
    * `FilterExpr` which contains either `LogicalExpr` or `TextExpr`. only accepted in `filter()`

topk-js/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+impl From<i32> for VectorFieldIndexMetric {
+  fn from(metric: i32) -> Self {
+    match metric {
+      1 => VectorFieldIndexMetric::Cosine,
+      2 => VectorFieldIndexMetric::Euclidean,
+      3 => VectorFieldIndexMetric::DotProduct,
+      4 => VectorFieldIndexMetric::Hamming,
+      _ => VectorFieldIndexMetric::Cosine, // default
```
jerguslejko, line 35: we shouldn't default here but rather raise

topk-js/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+
+#[napi(string_enum)]
+pub enum VectorFieldIndexMetric {
+  Cosine,
+  Euclidean,
+  DotProduct,
+  Hamming,
+}
+
+impl From<i32> for VectorFieldIndexMetric {
```
jerguslejko, line 28: where is the `i32` coming from? Ideally we'd have an enum here

topk-js/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+  Semantic {
+    model: Option<String>,
+    embedding_type: EmbeddingDataType,
+  },
+}
+
+impl From<control::FieldIndex> for FieldIndex {
+  fn from(field_index: control::FieldIndex) -> Self {
+    match field_index.index.unwrap_or_else(|| {
+      control::field_index::Index::KeywordIndex(control::KeywordIndex { index_type: 0 })
```
jerguslejko, line 96: nit: we shouldn't be using `0` here, but rather an enum. now its hard to understand what option this is, it's also easy to typo this into a different value

topk-js/src/client.rs
```diff
@@ -0,0 +1,227 @@
...
+        control::field_type::DataType::Text(_) => DataType::Text,
+        control::field_type::DataType::Integer(_) => DataType::Integer,
+        control::field_type::DataType::Float(_) => DataType::Float,
+        control::field_type::DataType::Boolean(_) => DataType::Boolean,
+        control::field_type::DataType::F32Vector(_) => DataType::F32Vector,
+        control::field_type::DataType::U8Vector(_) => DataType::U8Vector,
+        control::field_type::DataType::BinaryVector(_) => DataType::BinaryVector,
+        control::field_type::DataType::Bytes(_) => DataType::Bytes,
+      },
+      None => DataType::Text, // Default to Text type if none specified
```
jerguslejko, line 218: we should raise here, `None` should basically never happen

topk-js/src/collection.rs
```diff
@@ -0,0 +1,75 @@
...
+    Ok(
+      docs
+        .into_iter()
+        .map(|d| DocumentWrapper::from(d).into())
+        .collect(),
+    )
+  }
+
+  #[napi]
+  pub async fn upsert(&self, docs: Vec<HashMap<String, Value>>) -> Result<i32> {
```
jerguslejko, line 40: nit: we return `lsn` as `i32` here and as `i64` on the method below

topk-js/src/document.rs
```diff
@@ -0,0 +1,227 @@
...
+  ) -> napi::Result<Self> {
+    let mut result: i32 = 0;
+    napi::sys::napi_typeof(env, napi_val, &mut result);
+
+    match result {
+      napi::sys::ValueType::napi_string => {
+        Ok(Value::String(String::from_napi_value(env, napi_val)?))
+      }
+      napi::sys::ValueType::napi_number => Ok(Value::F64(f64::from_napi_value(env, napi_val)?)),
+      _ => panic!("unsupported value type: {:?}", result),
```
jerguslejko, line 42: this probably needs more types

topk-js/src/document.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+impl ToNapiValue for Value {
+  unsafe fn to_napi_value(env: sys::napi_env, val: Self) -> Result<sys::napi_value, napi::Error> {
+    match val {
+      Value::String(s) => String::to_napi_value(env, s),
+      Value::F64(n) => f64::to_napi_value(env, n),
+      Value::Bool(b) => bool::to_napi_value(env, b),
+      Value::U32(n) => u32::to_napi_value(env, n),
+      Value::U64(n) => u32::to_napi_value(env, n as u32),
```
jerguslejko, line 54: this is a not a safe cast

topk-js/src/expr.rs
```diff
@@ -0,0 +1,343 @@
...
+        expr: Expression::LogicalExpression {
+          expr: expr.equals(&LogicalExpression::Literal { value }),
+        },
+      },
+      _ => unreachable!("eq can only be called on logical expressions"),
+    }
+  }
+
+  #[napi]
+  pub fn not_eq(&self, value: Value) -> Expr {
```
jerguslejko, line 77: nit: `neq`

topk-js/src/expr.rs
```diff
@@ -0,0 +1,343 @@
...
+            value: Value::F64(value),
+          }),
+        },
+      },
+      _ => unreachable!("add can only be called on logical expressions"),
+    }
+  }
+
+  #[napi]
+  pub fn subtract(&self, value: f64) -> Expr {
```
jerguslejko, line 159: nit: `sub`

topk-js/src/expr.rs
```diff
@@ -0,0 +1,343 @@
...
+            value: Value::F64(value),
+          }),
+        },
+      },
+      _ => unreachable!("subtract can only be called on logical expressions"),
+    }
+  }
+
+  #[napi]
+  pub fn multiply(&self, value: f64) -> Expr {
```
jerguslejko, line 173: nit: `mul`

topk-js/src/expr.rs
```diff
@@ -0,0 +1,343 @@
...
+            value: Value::F64(value),
+          }),
+        },
+      },
+      _ => unreachable!("multiply can only be called on logical expressions"),
+    }
+  }
+
+  #[napi]
+  pub fn divide(&self, value: f64) -> Expr {
```
jerguslejko, line 187: nit `div`

topk-js/src/expr.rs
```diff
@@ -0,0 +1,343 @@
...
+    match &self.expr {
+      Expression::TextExpression { expr } => {
+        if let Expression::TextExpression { expr: other_expr } = &other.expr {
+          Self {
+            expr: Expression::TextExpression {
+              expr: expr.and(other_expr),
+            },
+          }
+        } else {
+          unreachable!("and can only be called on text expressions")
```
jerguslejko, line 211: this type of "error state" will disappear when `Expr` is split into multiple types

topk-js/src/expr.rs
```diff
@@ -0,0 +1,343 @@
...
+  //   field: options.as_ref().map(|o| o.field.clone()),
+  //   weight: options.as_ref().map_or(1.0, |o| o.weight),
+  // };
+
+  // let text_expr = TextExpression::Terms {
+  //   all: options.is_none(), // If options is not passed, all will be true
+  //   terms: vec![term],
+  // };
+
+  Expr::create_match(token)
```
jerguslejko, line 303: looks like `options` are ignored here

topk-js/src/expr.rs
```diff
@@ -0,0 +1,343 @@
...
+
+#[napi(object)]
+#[derive(Debug, Clone)]
+pub struct MatchOptions {
+  pub field: String,
+  pub weight: f64,
+}
+
+#[napi(js_name = "match")]
+pub fn match_(token: String, options: Option<MatchOptions>) -> Expr {
```
jerguslejko, line 291: tip: you can either do `match_` or use rust's fancy syntax `r#match`

topk-js/src/logical_expr.rs
```diff
@@ -0,0 +1,186 @@
...
+  //         "Binary(left={}, op={:?}, right={})",
+  //         left.as_ref().get().to_string(),
+  //         op,
+  //         right.get().to_string()
+  //       )
+  //     }
+  //   }
+  // }
+
+  pub fn equals(&self, other: &LogicalExpression) -> LogicalExpression {
```
jerguslejko, line 83: nit: `eq`

topk-js/src/logical_expr.rs
```diff
@@ -0,0 +1,186 @@
...
+    LogicalExpression::Binary {
+      left: MyBox(Box::new(self.clone())),
+      op: BinaryOperator::Eq,
+      right: MyBox(Box::new(other.clone())),
+    }
+  }
+
+  pub fn not_equals(&self, other: &LogicalExpression) -> LogicalExpression {
```
jerguslejko, line 91: nit: `neq`

topk-js/src/my_box.rs
```diff
@@ -0,0 +1,34 @@
+use napi::bindgen_prelude::{FromNapiValue, ToNapiValue};
+
+#[derive(Debug, Clone)]
+pub struct MyBox<T>(pub Box<T>);
```
jerguslejko, line 4: nit: rename from `MyBox` to something like `NapiBox`

topk-js/src/my_vec.rs
```diff
@@ -0,0 +1,2 @@
+#[derive(Debug, Clone)]
+pub struct MyVec<T>(pub Vec<T>);
```
jerguslejko, line 2: q: where is this used?

topk-js/src/client.rs
petertoth: It's coming from:
```rs
topk_protos::control::v1::VectorIndex
pub metric: i32
```

topk-js/src/collection.rs
petertoth: Updated it to `i64`, however, `u64` is not supported by NAPI by default. Check out [numbers](https://napi.rs/docs/concepts/values#numbers) section.
`u64` is supported via their [`BigInt`](https://napi.rs/docs/concepts/values#bigint) value which is a custom type in Typescript and imo feels a bit off since javascript user would just expect a `number` type.

topk-js/src/document.rs
petertoth: I replaced it with:
```rs
Value::U64(n) => i64::to_napi_value(env, n.try_into().unwrap()),
```
According to rust compiler: 
```
you can convert a `u64` to an `i64` and panic if the converted value doesn't fit: `.try_into().unwrap()`
```

topk-js/src/my_vec.rs
petertoth: Sorry just a leftover from vibe coding

topk-js/src/expr.rs
petertoth: looks like napi is complaining:

Screenshot 2025-04-02 at 13 49 53

jerguslejko: freaking procedural macros!

topk-js/Cargo.toml
jerguslejko: one thing to watch out for I guess -- [this comment](https://github.com/napi-rs/napi-rs/issues/1493#issuecomment-2746275863) mentions windows compatibility in relation to napi 3 release, worth verifying

topk-js/src/client.rs
jerguslejko: ah I see. [in pyo3](https://pyo3.rs/v0.24.1/class.html?highlight=enum#simple-enums) there are "int enums" so that mapping is handled automatically.

for this impl, I'd do something like

```
impl From<i32> for VectorFieldIndexMetric {
  fn from(metric: i32) -> Self {
    match topk_protos::v1::control::VectorDistanceMetric::try_from(metric) {
      Ok(topk_protos::v1::control::VectorDistanceMetric::Cosine) => VectorFieldIndexMetric::Cosine,
      Ok(topk_protos::v1::control::VectorDistanceMetric::Euclidean) => {
        VectorFieldIndexMetric::Euclidean
      }
      Ok(topk_protos::v1::control::VectorDistanceMetric::DotProduct) => {
        VectorFieldIndexMetric::DotProduct
      }
      Ok(topk_protos::v1::control::VectorDistanceMetric::Hamming) => {
        VectorFieldIndexMetric::Hamming
      }
      _ => VectorFieldIndexMetric::Cosine, // default
    }
  }
}
```

so re-using the `int <-> enum` mapping from automatically generated protos

topk-js/src/collection.rs
jerguslejko: yeah, this sucks... it makes me think that the client actually doesn't care about the value, it only uses `lsn` as a "snapshot of database". we could encode the strings using https://docs.rs/bs58/latest/bs58/. its base-58 strings that look nicely and random. wdyt? that way we'd return a string here, same in python, same in rust

topk-js/src/document.rs
jerguslejko: shouldn't we map to `BigInt` here?

petertoth: I would avoid using `BigInt` in JS-land since it's exported as a custom class in javascript thus unfamiliar with users. Maybe there's a use-case for that but I believe in majority of cases any number from rust should just be a `number` type in javascript

topk-js/src/collection.rs
petertoth: no strong opinions about this, string makes my life easier so if it makes sense all across the board then why not

topk-js/src/control/field_spec.rs
```diff
@@ -0,0 +1,140 @@
...
+    #[napi]
+    pub fn optional(&self) -> Self {
+        Self {
+            required: false,
+            ..self.clone()
+        }
+    }
+
+    #[napi]
+    pub fn keyword_index(&self) -> Self {
```
jerguslejko, line 44: these shouldn't exist, the syntax is `f32Vector().index(vectorIndex({metric: "..."}))`

topk-js/src/control/field_spec.rs
```diff
@@ -0,0 +1,140 @@
...
+        self.index(FieldIndex::KeywordIndex)
+    }
+
+    #[napi]
+    pub fn vector_index(&self, metric: VectorDistanceMetric) -> Self {
+        self.index(FieldIndex::VectorIndex { metric })
+    }
+
+    #[napi]
+    pub fn semantic_index(&self, model: Option<String>) -> Self {
```
jerguslejko, line 54: `embedding_type` should be passed as arg too

topk-js/src/control/field_spec.rs
```diff
@@ -0,0 +1,140 @@
...
+                        topk_protos::v1::control::field_type::DataType::Integer(Default::default())
+                    }
+                    DataType::Float => {
+                        topk_protos::v1::control::field_type::DataType::Float(Default::default())
+                    }
+                    DataType::Boolean => {
+                        topk_protos::v1::control::field_type::DataType::Boolean(Default::default())
+                    }
+                    DataType::F32Vector { dimension } => {
+                        topk_protos::v1::control::field_type::DataType::F32Vector(
```
jerguslejko, line 97: there are helpers on `topk_protos::v1::control::field_type::DataType` so you can write

```rs
                        topk_protos::v1::control::field_type::DataType::f32_vector(dimension)
```

instead of

```rs
                        topk_protos::v1::control::field_type::DataType::F32Vector(
                            topk_protos::v1::control::FieldTypeF32Vector { dimension },
                        )
```

topk-js/src/control/field_spec.rs
```diff
@@ -0,0 +1,140 @@
...
+    ) -> Result<Self> {
+        let env_env = Env::from_raw(env);
+
+        let is_field_spec = {
+            let env_value = Unknown::from_napi_value(env, value)?;
+            FieldSpec::instance_of(env_env, env_value)?
+        };
+
+        if is_field_spec {
+            FieldSpec::from_napi_value(env, value)
```
jerguslejko, line 135: this looks like infinite recrusion?

topk-js/src/data/document.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl From<topk_protos::v1::data::Document> for Document {
+    fn from(doc: topk_protos::v1::data::Document) -> Self {
+        Document {
+            fields: doc.fields.into_iter().map(|(k, v)| (k, v.into())).collect(),
+        }
+    }
+}
+
+pub struct DocumentWrapper(pub topk_protos::v1::data::Document);
```
jerguslejko, line 25: nit: `NapiDocument`

jerguslejko: also the inner field shouldn't be `pub`

topk-js/tests/collections.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+    collectionsCreated.push(name);
+    return collection;
+  };
+
+  test('create collection', async () => {
+    const schema = {
+      title: text(),
+      title_embedding: f32Vector(1536)
+        .required()
+        .index(vectorIndex({ metric: VectorDistanceMetric.Euclidean })),
```
jerguslejko, line 37: instead of `VectorDistanceMetric.Euclidean`, we should enable literal types, so `{ metric: "euclidean" | "dot_product" | "etc" }` to make this more user-friendly and avoid having to import `VectorDistanceMetric`

topk-js/tests/collections.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+      summary: text()
+        .required()
+        .index(semanticIndex({ model: "dummy", embeddingType: EmbeddingDataType.Float32 })),
```
jerguslejko, line 40: same as above, let's use literal values for inputs

topk-js/tests/expressions.spec.ts
```diff
@@ -0,0 +1,103 @@
+import { field, literal } from '../query';
+
+describe('Expression Tests', () => {
+  describe('Query expressions with flexible expressions', () => {
+    test('addition operations', () => {
+      expect(field('a').add(1).expr).toEqual(field('a').add(literal(1)).expr);
+      // JS doesn't support operator overloading, so we can't test 1 + field('a')
```
jerguslejko, line 7: makes sense, can you only type this once at the top instead of repeating in every fn?

topk-js/tests/setup.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  scope(name: string) {
+    return `${this.scopePrefix}-${name}`;
+  }
+}
+
+export function newProjectContext() {
+  const TOPK_API_KEY = process.env.TOPK_API_KEY.split('\n')[0].trim();
+  // const TOPK_HOST = process.env.TOPK_HOST || 'topk.io';
```
jerguslejko, line 19: why is this one commented out?

topk-js/query.js
```diff
@@ -0,0 +1,9 @@
+const { query } = require("./index");
```
jerguslejko, line 1: wdyt about putting entrypoints in a subfolder? like `/topk-js/entrypoints/...` or some other name but not in the root? root seems cluttered now

topk-js/src/data/function_expr.rs
```diff
@@ -0,0 +1,118 @@
...
+    VectorScore { field: String, query: VectorQuery },
+    SemanticSimilarity { field: String, query: String },
+}
+
+#[napi(namespace = "query")]
+pub fn semantic_similarity(field: String, query: String) -> FunctionExpression {
+    FunctionExpression::SemanticSimilarity { field, query }
+}
+
+#[napi]
```
jerguslejko, line 19: does this need `namespace=query`?

topk-js/src/data/function_expr.rs
```diff
@@ -0,0 +1,118 @@
...
+            }
+            VectorQuery::U8 { vector } => topk_protos::v1::data::Vector::byte(vector),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum VectorDistanceQuery {
+    Vector { vector: Vector },
+    Binary { vector: BinaryVector },
```
jerguslejko, line 40: why is `Binary` a separate option here instead of doing `Vector { vector: Vector::Binary {} }` just like for the other cases?

topk-js/src/data/logical_expr.rs
```diff
@@ -0,0 +1,254 @@
...
+impl Into<topk_protos::v1::data::LogicalExpr> for LogicalExpression {
+    fn into(self) -> topk_protos::v1::data::LogicalExpr {
+        self.expr.into()
+    }
+}
+
+impl Into<topk_protos::v1::data::LogicalExpr> for LogicalExpressionUnion {
+    fn into(self) -> topk_protos::v1::data::LogicalExpr {
+        match self {
+            LogicalExpressionUnion::Null => unreachable!(),
```
jerguslejko, line 218: why unreachable?

topk-js/src/data/mod.rs
```diff
@@ -0,0 +1,14 @@
+pub mod binary_expr;
+pub mod document;
+pub mod filter_expr;
+pub mod flexible_expr;
+pub mod function_expr;
+pub mod logical_expr;
+pub mod napi_box;
+pub mod query;
+// pub mod scalar;
```
jerguslejko, line 9: why commented

topk-js/src/data/query.rs
```diff
@@ -0,0 +1,183 @@
...
+            model,
+            query,
+            fields: fields.unwrap_or_default(),
+            topk_multiple,
+        });
+
+        new_query
+    }
+
+    // TODO: Remove this
```
jerguslejko, line 87: todo

topk-js/src/data/function_expr.rs
petertoth: this will have cascade effect since then we would need to put under query namepace:
`FlexibleExpr -> Stage -> Query -> CollectionClient(since it returns query)`

topk-js/index.d.ts
```diff
@@ -0,0 +1,208 @@
+/* auto-generated by NAPI-RS */
```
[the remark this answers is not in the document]
petertoth: index should still export client right? `import { Client } from 'topk-js'`

petertoth:
Continues in https://github.com/fafolabs/topk-sdk/pull/31

PR #15, opened 2025-03-29
improve python tests

jerguslejko:
* adjust python tests to account for dummy embedder/reranked (by not asserting ordering)
* add `earthly --secret TOPK_API_KEY +test-py` entrypoint for easier test running
* add CI for python when PR is labeled with `ready`


jerguslejko:
merging even when failing, `.get()` hasn't been deployed to prod yet

PR #25, opened 2025-04-15
`contains` operator

jerguslejko:


jerguslejko:
Tests are failing bc they run against `elastica` which does not have `contains` operator released yet. In the near future, we will run these tests against a local instance of topk.

PR #31, opened 2025-04-18
[RFC] topk js

petertoth:
Ready for comments.

Based on https://github.com/fafolabs/topk-sdk/pull/8


.github/workflows/js.yml
```diff
@@ -0,0 +1,584 @@
...
+        with:
+          image: node:${{ matrix.node }}-bookworm-slim
+          options: '--platform linux/arm/v7 -v ${{ github.workspace }}:/build -w /build/topk-js -e TOPK_API_KEY'
+          run: |
+            set -e
+            apt-get update
+            apt-get install ca-certificates -y
+            yarn test
+            ls -la
+  # universal-macOS:
```
jerguslejko, line 505: we probably need this one?

.github/workflows/js.yml
```diff
@@ -0,0 +1,584 @@
...
+  #         path: ./topk-js/artifacts
+  #     - name: Combine binaries
+  #       run: yarn universal
+  #     - name: Upload artifact
+  #       uses: actions/upload-artifact@v4
+  #       with:
+  #         name: bindings-universal-apple-darwin
+  #         path: ./topk-js/${{ env.APP_NAME }}.*.node
+  #         if-no-files-found: error
+  # publish:
```
jerguslejko, line 538: todo: uncomment before merging

topk-js/index.d.ts
```diff
@@ -0,0 +1,195 @@
...
+  get(name: string): Promise<Collection>
+  delete(name: string): Promise<void>
+}
+
+export interface ClientConfig {
+  apiKey: string
+  region: string
+  host?: string
+  https?: boolean
+  headers?: Record<string, string>
```
jerguslejko, line 29: how are you using headers?

topk-js/index.d.ts
```diff
@@ -0,0 +1,195 @@
...
+  export type VectorUnion =
+    | { type: 'Float', values: Array<number> }
+    | { type: 'Byte', values: Array<number> }
+    | { type: 'Binary', values: Array<number> }
+}
+
+export declare namespace query {
+  export class LogicalExpression {
+    static create(expr: LogicalExpressionUnion): LogicalExpression
+    get expr(): LogicalExpressionUnion
```
jerguslejko, line 82: q: do we need `get expr()` to run equality checks only, or something else as well?

topk-js/index.d.ts
```diff
@@ -0,0 +1,195 @@
...
+  export function match(token: string, field?: string | undefined | null, weight?: number | undefined | null, all?: boolean | undefined | null): TextExpression
+  export function select(exprs: Record<string, LogicalExpression | FunctionExpression>): Query
+  export function semanticSimilarity(field: string, query: string): FunctionExpression
+  export type TextExpressionUnion =
+    | { type: 'Terms', all: boolean, terms: Array<Term> }
+    | { type: 'And', left: TextExpression, right: TextExpression }
+    | { type: 'Or', left: TextExpression, right: TextExpression }
+  export type UnaryOperator =  'not'|
+  'isnull'|
+  'isnotnull';
```
jerguslejko, line 143: nit: `isNull` and `isNotNull`

topk-js/lib/data.d.ts
```diff
@@ -0,0 +1,7 @@
+import { data } from "../index";
+
+export declare const Vector: typeof data.Vector;
+export declare const binary: typeof data.binary;
```
jerguslejko, line 4: lets call this `bytes`

topk-js/lib/query.d.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export declare const bm25Score: typeof query.bm25Score;
+export declare const field: typeof query.field;
+export declare const literal: typeof query.literal;
+export declare const match: typeof query.match;
+export declare const select: typeof query.select;
+export declare const semanticSimilarity: typeof query.semanticSimilarity;
+export declare const vectorDistance: typeof query.vectorDistance;
+export declare const LogicalExpression: typeof query.LogicalExpression;
+export declare type LogicalExpressionUnion =
```
jerguslejko, line 11: curious: do we need to export this type for outside use?

topk-js/package.json
```diff
@@ -0,0 +1,82 @@
...
+  "files": [
+    "lib",
+    "index.js",
+    "index.d.ts"
+  ],
+  "exports": {
+    ".": {
+      "require": "./index.js",
+      "import": "./index.js",
+      "types": "./index.d.ts"
```
jerguslejko, line 33: if we export the same thing from index and from schema/query/data, how will the DX be? won't the editor suggest imports from `index` first?

topk-js/src/client/collection.rs
```diff
@@ -0,0 +1,167 @@
...
+        &self,
+        lsn: Option<i64>,
+        consistency: Option<ConsistencyLevel>,
+    ) -> Result<i64> {
+        let query = Query::new().count();
+
+        let docs = self
+            .client
+            .collection(&self.collection)
+            .query(
```
jerguslejko, line 78: you can use `client.collection(...).count()` here and the entire `match doc.fields.get("_count")` block can go awauy

topk-js/src/data/scalar.rs
```diff
@@ -0,0 +1,95 @@
...
+
+impl ToNapiValue for Scalar {
+    unsafe fn to_napi_value(
+        env: napi::sys::napi_env,
+        val: Self,
+    ) -> napi::Result<napi::sys::napi_value> {
+        match val {
+            Scalar::Bool(b) => bool::to_napi_value(env, b),
+            Scalar::U32(u) => u32::to_napi_value(env, u),
+            Scalar::U64(u) => u32::to_napi_value(env, u as u32),
```
jerguslejko, line 55: worth adding a TODO here, `u as u32` is lossy so we might consider returning `BigInt` in the future

topk-js/src/data/scalar.rs
```diff
@@ -0,0 +1,95 @@
...
+        let mut value_type = 0;
+
+        napi_typeof(env, napi_val, &mut value_type);
+
+        match value_type {
+            napi::sys::ValueType::napi_boolean => {
+                Ok(Scalar::Bool(bool::from_napi_value(env, napi_val)?))
+            }
+            napi::sys::ValueType::napi_number => {
+                let is_integer = is_napi_integer(env, napi_val);
```
jerguslejko, line 79: personal preference:

```rust
if is_napi_integer(env, napi_val) {
    Ok(Scalar::I32(i32::from_napi_value(env, napi_val)?)) 
} else { 
    Ok(Scalar::F64(f64::from_napi_value(env, napi_val)?))
}
```

is easier to read for me

topk-js/src/data/scalar.rs
```diff
@@ -0,0 +1,95 @@
...
+
+                match is_integer {
+                    true => Ok(Scalar::I32(i32::from_napi_value(env, napi_val)?)),
+                    false => Ok(Scalar::F64(f64::from_napi_value(env, napi_val)?)),
+                }
+            }
+            napi::sys::ValueType::napi_string => {
+                Ok(Scalar::String(String::from_napi_value(env, napi_val)?))
+            }
+            _ => Err(napi::Error::new(
```
jerguslejko, line 89: nit: tae the type and print it like `Invalid scalar type {:?}`

topk-js/src/data/utils.rs
```diff
@@ -0,0 +1,29 @@
+use napi::bindgen_prelude::*;
+
+pub unsafe fn is_napi_integer(env: napi::sys::napi_env, napi_val: napi::sys::napi_value) -> bool {
```
jerguslejko, line 3: I don't think this function is sound. if I have `[1.0, 2.0, 3.0]` it will consider all of these as ints

topk-js/src/data/value.rs
```diff
@@ -0,0 +1,244 @@
...
+pub enum Value {
+    String(String),
+    Bool(bool),
+    F64(f64),
+    U32(u32),
+    U64(u64),
+    I32(i32),
+    I64(i64),
+    F32(f32),
+    Binary(Vec<u8>),
```
jerguslejko, line 25: nit: lets call it `Bytes`

topk-js/src/data/value.rs
```diff
@@ -0,0 +1,244 @@
...
+    Vector(Vector),
+    Null,
+}
+
+#[napi(namespace = "data")]
+pub fn binary(values: Vec<u8>) -> Value {
```
jerguslejko, line 31: nit: let's call it `bytes` even though in proto we call it `Binary`

topk-js/src/data/value.rs
```diff
@@ -0,0 +1,244 @@
...
+    }
+}
+
+impl From<topk_protos::v1::data::Value> for Value {
+    fn from(value: topk_protos::v1::data::Value) -> Self {
+        match value.value {
+            Some(topk_protos::v1::data::value::Value::String(s)) => Value::String(s),
+            Some(topk_protos::v1::data::value::Value::F64(n)) => Value::F64(n),
+            Some(topk_protos::v1::data::value::Value::Bool(b)) => Value::Bool(b),
+            Some(topk_protos::v1::data::value::Value::U32(n)) => Value::I32(n.try_into().unwrap()),
```
jerguslejko, line 59: this will fail if number can't be converted. let's replace `.unwrap()` with `.expect()` so at least there's a good error message

topk-js/src/data/value.rs
```diff
@@ -0,0 +1,244 @@
...
+                }
+                Some(topk_protos::v1::data::vector::Vector::Byte(byte_vector)) => {
+                    Value::Vector(Vector::new(VectorUnion::Byte {
+                        values: byte_vector.values,
+                    }))
+                }
+                None => unreachable!("Invalid vector proto"),
+            },
+            Some(topk_protos::v1::data::value::Value::Null(_)) => Value::Null,
+            None => Value::Null,
```
jerguslejko, line 79: this is an invalid state, we should do `unreachable!("Invalid proto")` here instead

topk-js/src/data/vector.rs
```diff
@@ -0,0 +1,132 @@
...
+    unsafe fn from_napi_value(
+        env: napi::sys::napi_env,
+        value: napi::sys::napi_value,
+    ) -> napi::Result<Self> {
+        unsafe {
+            let mut is_array: bool = false;
+            napi::sys::napi_is_array(env, value, &mut is_array);
+
+            if is_array {
+                // If it's a JS array, convert to Vec<f64> and then to VectorDistanceQuery::Array
```
jerguslejko, line 63: nit: seems like the comment is outdated with `VectorDistanceQuery::Array`

topk-js/src/data/vector.rs
```diff
@@ -0,0 +1,132 @@
...
+    fn into(self) -> topk_protos::v1::data::Vector {
+        topk_protos::v1::data::Vector {
+            vector: Some(self.into()),
+        }
+    }
+}
+
+impl Into<topk_protos::v1::data::Value> for Vector {
+    fn into(self) -> topk_protos::v1::data::Value {
+        // let vector: topk_protos::v1::data::vector::Vector = self.into();
```
jerguslejko, line 127: nit: leftover

topk-js/src/expr/flexible.rs
```diff
@@ -0,0 +1,289 @@
...
+            FlexibleExpr::Expr(e) => LogicalExpression::to_napi_value(env, e),
+        }
+    }
+}
+
+impl Into<LogicalExpressionUnion> for FlexibleExpr {
+    fn into(self) -> LogicalExpressionUnion {
+        match self {
+            FlexibleExpr::String(s) => LogicalExpressionUnion::Literal {
+                value: Scalar::String(s),
```
jerguslejko, line 85: nit: you could probably do `value: s.into()`

topk-js/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#![deny(clippy::all)]
```
jerguslejko, line 1: why?

topk-js/src/query/query.rs
```diff
@@ -0,0 +1,157 @@
...
+            fields: fields.unwrap_or_default(),
+            topk_multiple,
+        });
+
+        new_query
+    }
+}
+
+#[napi(namespace = "query")]
+pub fn select(
```
jerguslejko, line 113: we should also export `filter()`, `top_k` and `count` query "starters", like we in python. `rerank()` on it's own doesn't make sense.

topk-js/src/schema/field_index.rs
```diff
@@ -0,0 +1,255 @@
...
+                    index: Some(FieldIndexUnion::VectorIndex {
+                        metric: topk_protos::v1::control::VectorDistanceMetric::try_from(v.metric)
+                            .expect("Unsupported vector distance metric")
+                            .into(),
+                    }),
+                },
+                topk_protos::v1::control::field_index::Index::SemanticIndex(s) => FieldIndex {
+                    index: Some(FieldIndexUnion::SemanticIndex {
+                        model: s.model,
+                        embedding_type: match s.embedding_type {
```
jerguslejko, line 216: nit: `s.embedding_type.map(|t|  topk_protos::v1::control::EmbeddingDataType::try_from(t).expect("Unsupported embedding data type").into())`

topk-js/src/schema/field_index.rs
```diff
@@ -0,0 +1,255 @@
...
+                topk_protos::v1::control::FieldIndex::vector(
+                    topk_protos::v1::control::VectorDistanceMetric::from(metric),
+                )
+            }
+            Some(FieldIndexUnion::SemanticIndex {
+                model,
+                embedding_type,
+            }) => topk_protos::v1::control::FieldIndex::semantic(
+                model,
+                embedding_type.map(|t| topk_protos::v1::control::EmbeddingDataType::from(t)),
```
jerguslejko, line 250: super nit: `map(topk_protos::v1::control::EmbeddingDataType::from)`

jerguslejko (commented):
leaving few comments, but looks veeeery good. let's pls address the outstanding items (most of them are small nits) and ship this thingg

jerguslejko:
one more think @petertoth we should add `topk-js` [here](https://github.com/fafolabs/topk-sdk/blob/57618ea5f263437aa1a338f51328d0f6e4ee6d20/Earthfile#L4C7-L4C7) and include it in gh actions [here](https://github.com/fafolabs/topk-sdk/blob/57618ea5f263437aa1a338f51328d0f6e4ee6d20/.github/workflows/pr.yml#L8C5-L8C5)

topk-js/index.d.ts
petertoth: just equality checks

topk-js/src/lib.rs
petertoth: it comes from napi-rs boilerplate

topk-js/index.d.ts
petertoth: removing it, but eventually we may want to allow users to add their custom headers

topk-js/src/data/utils.rs
jerguslejko: we checked and it is sound

jerguslejko (approved):
he FA, and the he FO. amazing work @petertoth!!

PR #35, opened 2025-04-21
Extract functional expressions into fn object | Extract query options into a single object

petertoth:
* extract functional expressions into a `fn` object in `topk-js/query` namespace similar to python:
  * from: `import { bm25Score } from 'topk-js/query'`
  * to: `import { fn } from 'topk-js/query'; fn.bm25Score()`
  
* extract query options, `lsn` and `consistencyLevel`, into a single struct which is passed into `query(), count(), get()` methods rather than spreading them as additional arguments
  * this makes it similar to python and makes it clearer in the code examples in the docs what exactly is being passed into these methods

  From(old):
  ```ts
  await client.collection("my_collection").query(
      ..., // query
	  2, // lsn
      "indexed" // Consistency Level: indexed
  )
  ``` 

  To(new):
  
  ```ts
  await client.collection("my_collection").query(
      ..., // query
     { lsn: 2, consistency: "indexed" }
  )
  ```

  Python(just for reference):
  
    ```py
    client.collection("my_collection").query(
        ..., # query
        consistency="indexed"
    )
    ```

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -82,8 +82,8 @@ test-js:
 
     # build
     WORKDIR /sdk/topk-js
-    RUN --mount=type=cache,target=/root/.yarn YARN_CACHE_FOLDER=/root/.yarn yarn install
-
+    ENV YARN_CACHE_FOLDER=/root/.yarn
+    RUN --mount=type=cache,target=/root/.yarn yarn install
 
     RUN --mount=type=cache,target=/usr/local/cargo/registry \
         --mount=type=cache,target=/usr/local/cargo/git \
--- a/topk-js/index.d.ts
+++ b/topk-js/index.d.ts
@@ -7,9 +7,9 @@ export declare class Client {
 }
 
 export declare class CollectionClient {
-  get(id: string, fields?: Array<string> | undefined | null, options?: ConsistencyOptions | undefined | null): Promise<Record<string, any>>
-  count(options?: ConsistencyOptions | undefined | null): Promise<number>
-  query(query: query.Query, options?: ConsistencyOptions | undefined | null): Promise<Array<Record<string, any>>>
+  get(id: string, fields?: Array<string> | undefined | null, options?: QueryOptions | undefined | null): Promise<Record<string, any>>
+  count(options?: QueryOptions | undefined | null): Promise<number>
+  query(query: query.Query, options?: QueryOptions | undefined | null): Promise<Array<Record<string, any>>>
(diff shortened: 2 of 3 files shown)


Earthfile
```diff
@@ -82,11 +82,10 @@ test-js:
 
     # build
     WORKDIR /sdk/topk-js
-    RUN --mount=type=cache,target=/usr/local/share/.cache/yarn/v6 \
-        yarn install
+    RUN --mount=type=cache,target=/root/.yarn YARN_CACHE_FOLDER=/root/.yarn yarn install
```
jerguslejko, line 85: nit: put `ENV YARN_CACHE_FOLDER=/root/.yarn` on the live above

Earthfile
```diff
@@ -82,11 +82,10 @@ test-js:
...
 
-    RUN --mount=type=cache,target=target \
```
jerguslejko, line 88: ideally we'd keep this one since that's the actual cache for rust complication

topk-js/index.d.ts
```diff
@@ -45,6 +45,11 @@ export interface CollectionFieldSpec {
 export type ConsistencyLevel =  'indexed'|
 'strong';
 
+export interface ConsistencyOptions {
```
jerguslejko, line 48: could we call this `QueryOptions`? all `get`, `count` and `query` are queries

topk-js/src/client/collection.rs
```diff
@@ -41,18 +48,20 @@ impl CollectionClient {
...
         id: String,
         fields: Option<Vec<String>>,
-        lsn: Option<i64>,
-        consistency: Option<ConsistencyLevel>,
+        options: Option<ConsistencyOptions>,
     ) -> Result<HashMap<String, Value>> {
+        let (lsn, consistency) = match &options {
+            Some(o) => (
+                o.lsn.map(|l| l as u64),
+                o.consistency.clone().map(|c| c.into()),
```
jerguslejko, line 56: nit: do we need the clone here?

jerguslejko (approved):
lgtm. just nits

Earthfile
petertoth: I presume rust code changes in `topk-js` and running `yarn build` will produce different contents of `target`. Therefore we want to cache `target` only when we are not changing the rust code right?

petertoth: Created an issue for this: https://github.com/fafolabs/topk-sdk/issues/36

PR #41, opened 2025-04-22
drop `top_k()` and `count()` query builders

jerguslejko:
It makes sense to let users start queries with `select()` or `filter()`. 

* `top_k()` on its own is not very useful as you'd only get the `_id` field back. forcing user to use `select()` makes it explicit what data is going to be returned
* `client.collection().query(count())` can be written as `client.collection().count()` so no need to export `count()` too

### DIFF
--- a/topk-js/index.d.ts
+++ b/topk-js/index.d.ts
@@ -123,7 +123,6 @@ export declare namespace query {
   'mul'|
   'div';
   export function bm25Score(): FunctionExpression
-  export function count(): Query
   export function field(name: string): LogicalExpression
   export function filter(expr: LogicalExpression | TextExpression): Query
   export type FunctionExpression =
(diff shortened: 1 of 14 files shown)


petertoth:
Logic LGTM. In JS, we also need to:

1. Run `yarn dev` inside `topk-js` to generate new `index.js` and `index.d.ts`
2. Go to `topk-js/lib/query.js` and remove lines:
    ```javascript
    module.exports.count = query.count;
    module.exports.topk = query.topk;
    ```
3. Go to `topk-js/lib/query.d.ts` and remove lines:
      ```typescript
      export declare const count: typeof query.count;
      export declare const topk: typeof query.topk;
      ```
4. Remove lingering tests(the ones which are testing `topk` and `count` query builders)

PR #48, opened 2025-04-25
rename top_k in py and rs -> topk 

petertoth:
Fixes https://github.com/fafolabs/topk-sdk/issues/40

* rename `Query.top_k` method to `Query.topk` in py sdk
* rename `Query.top_k` method to `Query.topk` in rs sdk
* update tests
* update docs
* bump sdks versions

### DIFF
--- a/topk-py/src/query/query.rs
+++ b/topk-py/src/query/query.rs
@@ -105,7 +105,7 @@ impl Query {
     }
 
     #[pyo3(signature = (expr, k, asc=false))]
-    pub fn top_k(&self, expr: LogicalExpr, k: u64, asc: bool) -> PyResult<Self> {
+    pub fn topk(&self, expr: LogicalExpr, k: u64, asc: bool) -> PyResult<Self> {
         Ok(Self {
             stages: [
                 self.stages.clone(),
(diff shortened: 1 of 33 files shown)


topk-rs/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "topk-rs"
-version = "0.1.0"
+version = "0.1.1"
```
petertoth, line 3: is this how we do it?

topk-py/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "topk-py"
-version = "0.1.19"
+version = "0.1.20"
```
petertoth, line 3: is this how we do it?

topk-rs/Cargo.toml
jerguslejko: we're not publishing topk-rs yet so yeah this is fine, doesn't have any effect now

topk-py/src/query/query.rs
```diff
@@ -105,7 +105,7 @@ impl Query {
     }
 
     #[pyo3(signature = (expr, k, asc=false))]
-    pub fn top_k(&self, expr: LogicalExpr, k: u64, asc: bool) -> PyResult<Self> {
+    pub fn topk(&self, expr: LogicalExpr, k: u64, asc: bool) -> PyResult<Self> {
```
jerguslejko, line 108: nit: let's add a (deprecated) alias for now, so existing use cases still work, we can drop it later.

jerguslejko (approved):
lgtm, we should release `0.2`

jerguslejko:
ran tests locally, merging

PR #49, opened 2025-04-25
topk-js docs: add ssr config information

jerguslejko:


petertoth (commented):
This is rather specific for using `topk-js` inside the backend(SSR) of web frameworks but I find it important we let our users know how they can make topk work in Vite or Webpack environments.
Without externalizing `topk-js` dependency the app will simply crash when hitting the SSR route

jerguslejko (commented):
noice

PR #52, opened 2025-04-25
[publish] remove publish on tag

jerguslejko:
Since this is a polyglot repo and we decided that we do not want a "bundled" (all sdks at once) release strategy, I'm removing GH triggers (`on: push: tags`). Each workflow can be dispatched manually from GH UI.

Also bumping python version.


jerguslejko:
@petertoth I'm gonna merge to unblock myself but please feel free to comment.

PR #57, opened 2025-04-26
change `collection(...).get(id: str)` to `collection(...).get(ids: list[str])`

jerguslejko:
to allow multiple to get multiple ids at the same time


jerguslejko:
ran tests locally (since there is a breaking change)

PR #64, opened 2025-04-28
change `lsn` from `u64` to `string`

jerguslejko:
javascript doesn't have u64s, so `lsn: u64` was problematic. simply encoding this value as str works in all languages (client doesn't manipulate it, just passes along to `query()`/`count()`)


jerguslejko:
ran tests locally works

PR #68, opened 2025-04-28
Add is null and is not null operators

### LINKED ISSUE #59: expose `is_null` and `is_not_null` operators
we have support for `is_null` and `is_not_null` in the data types ([py](https://github.com/fafolabs/topk-sdk/blob/53bb7b4cbcba984eec05740477f1dc757b4f663d/topk-py/src/expr/logical.rs#L9C2-L9C2), [js](https://github.com/fafolabs/topk-sdk/blob/53bb7b4cbcba984eec05740477f1dc757b4f663d/topk-js/src/expr/logical.rs#L369C8-L369C8), [rs](https://github.com/fafolabs/topk-sdk/blob/53bb7b4cbcba984eec05740477f1dc757b4f663d/topk-rs/src/expr/logical.rs#L4C20-L4C20)), but we don't expose the operators. I recently added the [not](https://github.com/fafolabs/topk-sdk/pull/53) operator, this should be a PR along the same lines

petertoth:
Fixes #59

### DIFF
--- a/topk-py/src/expr/logical.rs
+++ b/topk-py/src/expr/logical.rs
@@ -157,7 +157,19 @@ impl LogicalExpr {
 
     // Unary operators
 
-    // TODO: add `is_null()` and `is_not_null()` operators.
+    fn is_null(&self, py: Python<'_>) -> PyResult<Self> {
+        Ok(Self::Unary {
+            op: UnaryOperator::IsNull,
+            expr: Py::new(py, self.clone())?,
+        })
+    }
+
+    fn is_not_null(&self, py: Python<'_>) -> PyResult<Self> {
+        Ok(Self::Unary {
+            op: UnaryOperator::IsNotNull,
+            expr: Py::new(py, self.clone())?,
+        })
+    }
 
     // Comparison operators
 
--- a/topk-rs/src/expr/logical.rs
+++ b/topk-rs/src/expr/logical.rs
@@ -109,12 +109,24 @@ impl LogicalExpr {
         }
     }
 
-    // Unary operators
-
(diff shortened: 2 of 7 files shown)


topk-py/src/expr/logical.rs
```diff
@@ -157,7 +157,19 @@ impl LogicalExpr {
 
     // Unary operators
 
-    // TODO: add `is_null()` and `is_not_null()` operators.
+    fn is_null(&self, py: Python<'_>) -> PyResult<Self> {
```
jerguslejko, line 160: we can take ownership of `self` here to avoid the `.clone()` below

topk-rs/src/expr/logical.rs
```diff
@@ -109,12 +109,24 @@ impl LogicalExpr {
...
 
-    // Unary operators
-
     // NOTE: we don't expose `.not()` operator on the fluent query builder.
     // Instead, we use the `not()` query builder from the `query` module.
 
-    // TODO: add `is_null()` and `is_not_null()` operators.
+    // Unary operators
+
+    pub fn is_null(&self) -> Self {
```
jerguslejko, line 117: same `&self` -> `self` here. we should do this for all operators (separate pr?)

jerguslejko (approved):
lets tweak the ownership otherwise lgtm, tests are passing now (fixed is_null in https://github.com/topk-io/ddb/pull/469)

topk-rs/src/expr/logical.rs
jerguslejko: eh, the `to_owner()` call is going to clone anyway.. fine to keep as is

PR #78, opened 2025-05-12
[wip] Documentation refactoring

petertoth:


docs/concepts/vector-search.mdx
```diff
@@ -0,0 +1,157 @@
...
+3. Lastly, we return the top 10 results by the `title_similarity` field in descending order.
+
+## Combinding vector search with metadata filtering
+
+While vector search captures semantic similarity, it often retrieves results that may not satisfy all user or system-level constraints.
+
+**Metadata filtering** is used to enforce hard constraints based on crieteria such as
+category, time range, price range, availability, user permissions, language, or geographic region.
+
+In a multi-tenant setting, metadata filtering is used to narrow down the search space for a specific tenant by filtering documents based on the given tenant ID.
```
jerguslejko, line 119: imo this is out of place in "vector search", I'd remove this from this page

docs/concepts/multi-vector-search.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+Let's define a collection with two [semantic indexes](/collections/create#semantic-index):
+
+<CodeGroup>
+
+```python Python
+from topk_sdk.schema import text, vector, semantic_index
+
+client.collections().create(
+    "papers",
```
jerguslejko, line 19: this should be call `paper_paragraphs` or something (maybe article instead of paper?) since its a collection of paragraphs not paers

docs/concepts/multi-vector-search.mdx
```diff
@@ -0,0 +1,127 @@
...
+];
+```
+
+</CodeGroup>
+
+Let's break down the above example:
+
+1. First, we calculate semantic similarity between the query and the full paper summary as well as the paragraph summary.
+2. Then, we blend the two scores by giving 70% weight to the full paper and 30% weight to the paragraph.
+3. Finally, we retrieve the top 10 results ranked by the blended score.
```
jerguslejko, line 127: we should add a bit callout section either at the end or at the beginning. NO ONE LETS YOU DO THIS BUT TOPK. HAVE FUN IN YOUR PINECONES

docs/concepts/true-hybrid-search.mdx
```diff
@@ -0,0 +1,266 @@
+---
+title: "True Hybrid search"
```
jerguslejko, line 2: imo this isn't really a guide since both semantic search guide and keyword search guide show hybrid. this is page is in the same category as architecture or similar. we should highlight it more

docs/concepts/true-hybrid-search.mdx
```diff
@@ -0,0 +1,266 @@
...
+True Hybrid Search allows you to combine multiple retrieval techniques — such as multi-vector search, hybrid (vector \+ text) search
+sort the results using a custom scoring function **within a single query**.
+
+## How TopK differs from other "hybrid" search systems
+
+Most databases that offer "hybrid" search maintain separate vector and keyword indexes. When a query is executed:
+
+1. They run the query against both indexes separately
+2. After, each index returns its top results (e.g. first 100 candidates)
+3. Lastly, use techniques like Reciprocal Rank Fusion (RRF) to merge and rerank these two sets of results
```
jerguslejko, line 14: nit, I wouldnt write this as "after" and "lastly" like its a mini story.

```
This is how the WBA (wanna-be-algorithm) works:

1. Run two separate queries against respective indexes
2. Each query returns its top results (e.g. first 100 candidates)
3. Use techniques like Reciprocal Rank Fusion (RRF) to merge and rerank these two sets of results
```

check with @marekgalovic for the working for (3). rrf is only one of many afaik

docs/concepts/true-hybrid-search.mdx
```diff
@@ -0,0 +1,266 @@
...
+
+<br />
+
+```mermaid
```
jerguslejko, line 18: question: can we "center" these? looks off when they are on the left

docs/concepts/true-hybrid-search.mdx
```diff
@@ -0,0 +1,266 @@
...
+
+  D --> F{Merge and rerank top 200 results using RRF}
+  E --> F
+
+  F --> G@{ shape: stadium, label: "Final 100 ranked results" }
+```
+
+<br />
+
+This approach is fundamentally **probabilistic** - the final "top 100" results are not guaranteed to be the actual best matches because some potential matches might be missed if they don't appear in either index's top results.
```
jerguslejko, line 34: check with @marekgalovic for wording here

docs/concepts/true-hybrid-search.mdx
```diff
@@ -0,0 +1,266 @@
...
+
+**TopK is different.** We scan both vector and keyword indexes simultaneously within a single query, ensuring that our "top 100" results are the **actual** top 100 - not just a probabilistic approximation:
+
+```mermaid
+flowchart TD
+    A[User Query]
+    A --> B[(Keyword Index)]
+    A --> C[(Vector Index)]
+
+    B --> D{Simultaneously scan both indexes and retrieve top 100 results}
```
jerguslejko, line 44: we don't scan "both" indexes. we have one index

docs/concepts/true-hybrid-search.mdx
```diff
@@ -0,0 +1,266 @@
...
+    C --> D
+    D --> E@{ shape: docs, label: "Combined top 100 Keyword and Vector search results" }
```
jerguslejko, line 46: its not "combined" we only have one index

docs/concepts/true-hybrid-search.mdx
```diff
@@ -0,0 +1,266 @@
...
+    E --> F@{ shape: stadium, label: "Final 100 ranked results" }
```
jerguslejko, line 47: I wouldn't say "Final" here as it implies this wasn't the case all along. something like "True/Real/Actual top 100 results"

docs/concepts/true-hybrid-search.mdx
```diff
@@ -0,0 +1,266 @@
...
+With TopK, you can:
+
+- Retrieve documents based on multiple embeddings -- **multi-vector retrieval**
+- Combine semantic similarity (or vector distance) with keyword search -- **true hybrid retrieval**
+- Filter documents to enforce hard constraints -- **metadata filtering**
+- Apply custom scoring functions blending multiple ranking factors -- **custom scoring**
+
+
+
+## Implementing Hybrid Search(Vector + Keyword)
```
jerguslejko, line 60: nit space

docs/concepts/true-hybrid-search.mdx
jerguslejko: to really nail this point, we should add actual comparisons to other dbs and call them out for being imposters. for example, pinecone only has hybrid in the python client (bc its implemented client-side). tpuf requires you to write bunch of code.

since this is a differentiator, we need to actaully convince the user topk is better

docs/concepts/true-hybrid-search.mdx
```diff
@@ -0,0 +1,266 @@
...
+```python Python
+from topk_sdk.query import select, fn, match
+
+docs = client.collection("articles").query(
+    select(
+        "title",
+        content_similarity=fn.semantic_similarity("content", "climate change policies"),
+        text_score=fn.bm25_score()
+    )
+    .filter(match("carbon tax") | match("renewable energy"))  # Ensure keyword relevance
```
jerguslejko, line 105: btw: matching on "carbon tax" will match on "carbon" and then "tax". matching on "tax" is somewhat more generic. I think what you intuitively wanted to do here is "phrase search" -- which we don't support just yet (or ever?)

docs/concepts/true-hybrid-search.mdx
```diff
@@ -0,0 +1,266 @@
...
+      10
+    )
+);
+```
+
+</CodeGroup>
+
+Let's break down the above example:
+
+- We retrieve documents based on semantic meaning (`content_score`) and keyword matching (`keyword_match`).
```
jerguslejko, line 135: `content_score` is a hallucination?

docs/concepts/reranking.mdx
```diff
@@ -0,0 +1,64 @@
+---
+title: "Reranking"
+---
+
+Vector search or keyword-based functions (like BM25 or ANN) are optimized for speed, not perfect accuracy.
```
jerguslejko, line 5: what is the point you are trying to make here? if its "you can make shit better by calling `rerank()`" I wouldn't do it by saying that bm25/ann are not perfect. that's kind of shitting on ourselves. bm25 doesn't trade recall of speed, ANN does.

docs/concepts/reranking.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+<ParamField query="model" type="string" default="cohere/rerank-v3.5">
+  The model to use for reranking. Currently, only [`cohere/rerank-v3.5`](https://cohere.com/blog/rerank-3pt5) is supported.
+</ParamField>
+
+<ParamField query="query" type="string" optional>
+  The query text to rerank against. Uses arguments from `semantic_similarity()` function if not specified.
+</ParamField>
+
+<ParamField query="fields" type="string[]" optional>
```
jerguslejko, line 27: I think this only works if you have a single `semantic_similarity()` call. if you have multiple, you must provide the query. but we should check, not sure

docs/concepts/consistency.mdx
```diff
@@ -9,7 +9,7 @@ TopK supports three different consistency levels, allowing you to choose the rig
 TopK follows a write-ahead log (WAL) and compaction-based architecture:
```
jerguslejko, line 9: imo consistency shouldnt be in "Guides" but in some other more highlevel section along with architecture

docs/concepts/consistency.mdx
```diff
@@ -9,7 +9,7 @@ TopK supports three different consistency levels, allowing you to choose the rig
...
 
 1. **Writer** appends incoming data to the **WAL**.
-2. **Compactor** processes the WAL and produces optimized, indexed files.
+2. **Compactor** processes the WAL and produces optimized, indexed files, also known as *compacted files*.
```
jerguslejko, line 12: a mini diagram would be nice to show the data flow. like here https://www.topk.io/resources/engineering-journal

docs/concepts/consistency.mdx
jerguslejko: (cant comment below) in `Indexed Consistency` we can `It is best suited for analytical workloads and batch processing where absolute real-time consistency isn’t required.`. not sure that's right check with @marekgalovic

docs/concepts/multi-tenancy.mdx
```diff
@@ -0,0 +1,55 @@
...
+```typescript Javascript
+await client.collection("books").upsert([
+  { _id: "tenant-1234_1", title: "The Great Gatsby", author: "F. Scott Fitzgerald" },
+  { _id: "tenant-1234_2", title: "To Kill a Mockingbird", author: "Harper Lee" },
+]);
+```
+
+</CodeGroup>
+
+### Querying documents for a specific tenant
```
jerguslejko, line 31: we should explain more how this is different from other solutions and how performance improves with this ferature

docs/docs.json
```diff
@@ -17,23 +17,28 @@
...
+              "concepts/multi-vector-search",
+              "concepts/keyword-search",
+              "concepts/true-hybrid-search",
+              "concepts/reranking",
+              "concepts/consistency",
+              "concepts/multi-tenancy"
             ]
           },
           {
             "group": "Documents",
```
jerguslejko, line 38: nit: I'd call it `Document API` and `Collection API`

docs/documents/upsert.mdx
```diff
@@ -67,35 +65,204 @@ await client.collection("books").upsert([
...
+    // float_embedding: f32Vector([0.12, 0.67, 0.82, 0.53]),
+    u8_embedding: u8Vector([0, 1, 2, 3]),
+    binary_embedding: binaryVector([0, 1, 1, 0]),
+    bytes: bytes([0, 1, 1, 0]),
+  },
+]);
+```
+</CodeGroup>
+
+## Helper functions
```
jerguslejko, line 194: we should explain why these are needed to disambiguate between common types `number[]`

docs/documents/query.mdx
```diff
@@ -1,53 +1,125 @@
 ---
 title: "Query documents"
-description: "Use the `query` method to search for documents."
+description: "Use the `query` function to search for documents."
 ---
 
 import Support from "/snippets/support.mdx";
 
-TopK allows you to query documents using a SQL-inspired syntax. It comes with **semantic search**, **text search**, **vector search**, and **filtering capabilities** out of the box.
+TopK provides a SQL-inspired syntax for querying documents. It features built-in **semantic search**, **text search**, **vector search**, metadata **filtering** as well as **reranking** capabilities.
```
jerguslejko, line 8: instead of `SQL-inspired` we can call it "data frame-like" so people know its similar to numpy/etc

jerguslejko: this also reminds me, tpuf and other have helper methods to convert their result collection into numpy array directly. I wouldn't pollute our api with this bs, but we can mention/show it in docs

docs/documents/query.mdx
```diff
@@ -1,53 +1,125 @@
...
+  * these fields can be used in stages such as Filter, TopK or Rerank
+* **Filter stage** - Filter the documents that will be returned in the query results
+  * filters can be applied to static fields, computed fields such as `vector_distance()` or `semantic_similarity()` or custom properties computed inside `select()`
+* **TopK stage** - Return the top `k` results based on the provided logical expression
+* **Count stage** - Return the total number of documents matching the query
+* **Rerank stage** - Rerank the results
+
+All queries **must** have either TopK or Count collection stage.
+
+<Tip>
```
jerguslejko, line 26: you can't have multiple topk or count

docs/documents/query.mdx
```diff
@@ -1,53 +1,125 @@
...
 <CodeGroup>
 
 ```python Python
-from topk_sdk.query import select, fn, match
+from topk_sdk.query import select, field
+
+client.collection("books").query(
+  select(
+    title=field("title"),
+    "published_year" # you can also select the static fields directly
```
jerguslejko, line 56: in python positional args must come before keyword args, this doesn't compile

docs/documents/query.mdx
```diff
@@ -1,53 +1,125 @@
...
+  select({
+    title: field("title"),
+  })
+  ...
+)
+```
+
+</CodeGroup>
+
+### Select expressions
```
jerguslejko, line 75: this is the same as the example above?

jerguslejko: probably a mistake and should show `fn.` usage

jerguslejko (approved):
left bunch of comments, it looks good! lets fix the little mistakes and ship itttt

docs/concepts/semantic-search.mdx
```diff
@@ -62,6 +65,25 @@ docs = client.collection("books").query(
...
 )
+
+# Example results:
+
+[
+  {
+    "_id": '2',
+    "title": 'The Catcher in the Rye',
+    "title_similarity": 0.9497610926628113,
+    "_rank": 0
```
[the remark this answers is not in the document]
petertoth: These fields are added by the `.rerank()` function. Not sure if we should be exposing them but in fact they are

docs/concepts/consistency.mdx
petertoth: I plan to add "Architecture" section in the next iteration along with more "under-the-hood" content. Taking it out into a custom standalone section now would feel a bit off. I am not against renaming the existing section to "Concepts" again but I feel like "Guides" is more common in docs world.

docs/quickstart.mdx
```diff
@@ -176,6 +125,8 @@ results = client.collection("books").query(
     )
     # Sort results by the `title_similarity` field, selecting the top 10 results
```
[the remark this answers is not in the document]
petertoth: What about:

```
# Sort results by the `title_similarity` field which is defined "on the fly",
# selecting the top 10 results
```
?

docs/quickstart.mdx
```diff
@@ -194,14 +145,16 @@ const results = await client.collection("books").query(
   })
     // Sort results by the `title_similarity` field, selecting the top 10 results
     .topk(field("title_similarity"), 10)
+    // Rerank results using cohere/rerank-v3.5 model
```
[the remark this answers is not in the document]
petertoth: `Rerank results using TopK's built-in reranking feature` ?

docs/concepts/vector-search.mdx
```diff
@@ -0,0 +1,157 @@
...
+title: "Vector search"
+---
+
+Vector search is the essential feature of TopK. With that in mind, TopK is designed to:
+
+* Handle **billion-scale vector search** efficiently across massive datasets
+* Achieve **> 95% recall** in vector search — consistently retrieving more than 95% of the true top-K nearest neighbors on practical workloads
+* Provide **low latency** (< 100ms) for any query across any size of dataset
+
+## How to perform a vector search
```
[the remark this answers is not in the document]
petertoth: I can think of these alternatives:
1. Execute semantic search 
2. Run semantic search
3. (When describing logic) Find relevant results based on the semantic meaning of title values

Is the goal to eliminate use the of "perform ... search" altogether or just switch phrasing here and there to reduce the number of "perform ... search" usage ?

docs/quickstart.mdx
```diff
@@ -159,9 +108,9 @@ await client.collection("books").upsert([
 
 </CodeGroup>
 
-## 5. Query your collection
+## 3. Query your collection
 
-Now, send your first **semantic search query**:
+Now, perform your first **semantic search** query:
```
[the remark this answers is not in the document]
petertoth: Since `fn.semanticSimilarity()` does not really "find" or "filter" documents, it just adds similarity score to each document for a given query, what about:

`fn.semantic_similarity() scores documents by how semantically similar their titles are to "classic American novel".` ?

docs/concepts/semantic-search.mdx
```diff
@@ -241,11 +287,11 @@ const docs = await client.collection("books").query(
 );
 ```
 
-</CodeGroup>
+</CodeGroup> */}
 
 ### Lexical scoring with reranking
```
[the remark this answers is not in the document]
petertoth: Hmm when thinking about it, this section should probably be moved to the "Reranking" page and the value displayed should be something like: "Search by keywords and sprinkle some reranking AI to push the most relevant results upfront"

docs/concepts/vector-search.mdx
```diff
@@ -0,0 +1,157 @@
+---
+title: "Vector search"
+---
+
+Vector search is the essential feature of TopK. With that in mind, TopK is designed to:
+
+* Handle **billion-scale vector search** efficiently across massive datasets
+* Achieve **> 95% recall** in vector search — consistently retrieving more than 95% of the true top-K nearest neighbors on practical workloads
```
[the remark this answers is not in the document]
petertoth: we discussed this over lunch and he said >~ 95 but important to add "on practical workloads"

docs/concepts/true-hybrid-search.mdx
petertoth: yes, need to use custom css

docs/concepts/true-hybrid-search.mdx
petertoth: changed it to just `"carbon"` for now

docs/documents/query.mdx
petertoth: The goal was to provide docs for both `select()`(query builder) as well as `field()` - select expression.
But I see your point that both examples look almost identical. Perhaps I should not use `field()` expr inside the `select()` docs

docs/quickstart.mdx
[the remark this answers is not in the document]
jerguslejko:
```
# Sort results by `title_similarity` (computed in `select()` above), and limit to top 10 results
```

docs/concepts/semantic-search.mdx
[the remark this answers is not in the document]
jerguslejko: ugh, I don't this we should be? ping @marekgalovic but that's a bug possibly

docs/concepts/vector-search.mdx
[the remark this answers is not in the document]
jerguslejko: the latter, just so it sounds less repetetive

docs/documents/query.mdx
jerguslejko: see numpy/pandas/polars for py, and danfo.js/dataframe-js for js. imo "dataframe api" is a common abstraction for people working with data

docs/concepts/multi-vector-search.mdx
petertoth: What about this in the beginning:

`In TopK, you can run **multiple vector searches in a single query** and apply a **different ranking score** to individual vector search results.`

?

PR #84, opened 2025-05-15
Extract match options into an object

petertoth:
* Extract match options into an optional object

```js
interface MatchOptions {
  field?: string;
  weight?: number;
  all?: boolean;
}

match(token: string, options?: MatchOptions)
```

* tests
* update docs


jerguslejko (approved):
lgtm. this is a breaking change so let's bump the minor version

PR #136, opened 2025-06-25
inline `topk-protos` inside `topk-rs`

jerguslejko:
We want to get rid of `topk-protos` crate for two reasons:
* we cannot publish `topk-rs` as a crate to crates.io without publishing `topk-protos` crate as well. This is because `topk-rs` include `topk-protos` as a "path dependency" (eg. in Cargo.toml: `topk-protos = { path = "../topk-protos" }`). To work around this, we'd have to publish `topk-protos` too which is unnecessary.
* right now we have a mapping layer between `topk-protos` and `topk-rs` (just like we do for `topk-rs <> topk-py` and `topk-rs <> topk-js`) which is proving confusing (since we're using prost-generated stubs directly in rust anyway)

### DIFF
--- a/topk-py/topk_sdk/query/__init__.pyi
+++ b/topk-py/topk_sdk/query/__init__.pyi
@@ -99,7 +99,13 @@ class fn:
     @staticmethod
     def vector_distance(
         field: builtins.str,
-        vector: typing.Union[list[int], list[float], topk_sdk.data.Value],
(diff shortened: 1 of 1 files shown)


jerguslejko:
Tests are failing because elastica is not yet deployed with latest version. Tested locally, good to go.

PR #154, opened 2025-07-05
[ci] rename github runners

jerguslejko:
Runners `amd64/arm64` were renamed to `public-amd64/public-arm64` in our github org.


jerguslejko:
Merging since rust tests fail because target cluster was not yet released. This flow needs fixing.

PR #164, opened 2025-07-07
[py] pin maturin version

jerguslejko:
we already pin maturin version [here](https://github.com/topk-io/topk/blob/9c2e6cd1739439572f7a1f74f58f8ac3f163f161/Earthfile#L44C32-L44C32) so pinning in release wfs too to get consistent experience.

### DIFF
--- a/.github/workflows/publish_python.yml
+++ b/.github/workflows/publish_python.yml
@@ -49,6 +49,7 @@ jobs:
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
+          maturin-version: 1.8.7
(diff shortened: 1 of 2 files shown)


robertschonfeld (approved):
Currently we are using maturin action 1.9.0 and after https://github.com/topk-io/topk/pull/150 it seems to be ok. So maybe we are better off sticking to 1.9.0 and also changing here https://github.com/topk-io/topk/blob/4010545cf82730d9dd497ae939f77b5945119a08/Earthfile#L44

Otherwise thanks for doing this!

PR #186, opened 2025-08-06
vector -> list

marekgalovic:
Since we now have a generic `list` type, deprecate the `Vector` type to remove ambiguity.


petertoth (approved):
🔥

PR #190, opened 2025-08-07
Value everything 

petertoth:


topk-js/src/expr/function.rs
```diff
@@ -9,7 +11,7 @@ impl FunctionExpression {
         FunctionExpression(FunctionExpressionUnion::KeywordScore)
     }
 
-    pub(crate) fn vector_score(field: String, query: QueryVector) -> Self {
+    pub(crate) fn vector_score(field: String, query: Value) -> Self {
```
jerguslejko, line 14: nit: I would remove this entire impl and inline at the callsite

topk-py/src/query/mod.rs
```diff
@@ -127,20 +126,8 @@ pub fn bm25_score() -> FunctionExpr {
...
-            Vectorish::SparseF32(indices, values) => {
-                QueryVector::Sparse(SparseVector::F32 { indices, values })
-            }
-            Vectorish::SparseU8(indices, values) => {
-                QueryVector::Sparse(SparseVector::U8 { indices, values })
-            }
-        },
-    }
+pub fn vector_distance(field: String, query: Value) -> FunctionExpr {
+    FunctionExpr::VectorScore { field, query }
```
jerguslejko, line 130: we need to match on value types here too

jerguslejko (changes requested):

PR #204, opened 2025-08-27
Add skip_refine to Python type definitions

petertoth:
* add skip_refine to python definitions `topk__sdk/query/___init___.pyi`
* bump

### DIFF
--- a/topk-js/Cargo.toml
+++ b/topk-js/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "topk-js"
-version = "0.4.8"
+version = "0.4.9"
 edition = "2021"
 description = "TopK JavaScript SDK with TypeScript support"
 license-file = "LICENSE"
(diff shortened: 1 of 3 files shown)


topk-py/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "topk-py"
-version = "0.4.8"
+version = "0.4.9"
```
marekgalovic, line 3: Should we bump all SDKs to keep them in sync?

marekgalovic (approved):

topk-py/Cargo.toml
petertoth: actually yes let's do that

jerguslejko: yup, we should always bump all sdks. we should add a CI check that enforces this

PR #205, opened 2025-08-27
Add string list & f32_list constructors

petertoth:
* adds string list constructors
* adds missing TS types for vector_distance options (we need to run `yarn dev` or `yarn build` for `.d.ts` regeneration
* docs
* added missing TS types here and released the bump https://github.com/topk-io/topk/pull/208
* don't release wait for `contains`, `in_` and `any` `all` PR and then release minor


petertoth:
> @petertoth can we add a CI check to make sure types are always generated? something like`yarn dev --check`.
> 
> also, we should add a check confirming all version numbers are the same. the bump-version script works well for this, I'd just add a check command

Sure we should definitely do that. I think we already had this issue with types being not in sync in TS (and JS!)

topk-py/tests/test_data_list.py
```diff
@@ -96,3 +121,26 @@ def test_f64_list_invalid_input():
         TypeError, match="Expected list\\[float\\] for f64_list\\(\\) function"
     ):
         f64_list(123)  # type: ignore
+
+
+def test_string_list():
+    result = string_list(["1", "2", "3"])
+    assert result is not None
+    assert str(result) == "['1', '2', '3']"
```
robertschonfeld, line 129: In general, could we assert the actual value in these tests, e.g. result == ['1', '2', '3']?

I was thinking the idea with asserting the str representation here is for cases when we return a class (like Value or List) which may not be easy to construct in Python directly

robertschonfeld (approved):
LGTM, thank you.

However I recommend releasing this only after changing Value to not be converted to native py list as we discussed. Until then, all list data constructs in python are a "noop" and for that reason this PR also doesn't address the issue of upserting an empty string list into a keyword indexed field.

petertoth:
> @petertoth can we add a CI check to make sure types are always generated? something like`yarn dev --check`.
> 
> also, we should add a check confirming all version numbers are the same. the bump-version script works well for this, I'd just add a check command

Addressed in this PR: https://github.com/topk-io/topk/pull/206

PR #211, opened 2025-08-29
docs: List contains and in

petertoth:
* we shall merge #210 before and bump

### DIFF
--- a/docs/documents/query.mdx
+++ b/docs/documents/query.mdx
@@ -917,19 +917,82 @@ then be used to scope the query to a specific tenant.
 
 #### contains
 
-The `contains` operator can be used on string fields to match documents that include a specific substring. It is **case-sensitive** and is particularly useful in scenarios where you need to filter results based on a portion of a string.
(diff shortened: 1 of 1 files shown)


petertoth:
> Hey, a few points to consider adding
> 
> * `x CONTAINS y` works exactly the same as `y IN x`. Perhaps the docs should simply state this instead of repeating. The only reason we added 2 operators is so we can tell which one people are using
> * For the substring contains, I would ephasise the contrast to match. Match on text field uses our text pipeline i.e. tokenisation and stemming. Contains avoids this.
> * It should be mentioned somewhere in the docs that you can use a list of strings with a keyword index if you dont want to use our text pipeline but rather provide your own tokens. Here the contrast to match is that you can use contains even without a keyword index.

Thanks for the valuable feedback! Let me know if this makes it clearer: https://github.com/topk-io/topk/pull/211/commits/ef7c61cf73cff5d184f6e1e2c1e253acf6457bd5

PR #220, opened 2025-09-04
Add basic pre-commit

robertschonfeld:


jerguslejko:
@robertschonfeld LGTM, thanks for adding this.

bonus extension would be to run pre-commit in CI and block merges if not formatted

PR #224, opened 2025-09-05
feat: Async Collection Client

petertoth:
* install [pyo3-async-runtimes](https://github.com/PyO3/pyo3-async-runtimes) with `features = ["tokio-runtime"]`
* implement `AsyncCollectionClient`
* expose `AsyncCollectionClient` on `topk_client.async_collection("my-collection")`
* types
* tests
* docs

### DIFF
--- a/docs/documents/async-await.mdx
+++ b/docs/documents/async-await.mdx
@@ -45,6 +45,40 @@ async def main():
 asyncio.run(main())
 ```
 
+### Parallel Queries
+
+You can run multiple queries concurrently using `asyncio.gather()`:
+
+```python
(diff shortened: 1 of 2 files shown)


jerguslejko (changes requested):
the standard pattern when providing both sync & async methods in python SDK is to have a separate client class, so `Client` (for sync` and `AsyncClient` for async. 

In the async SDK, all methods must be async. In the current implementation, calling `client.collections().create()` (sync method) alongside async code (eg. `await client.async_collection().query()`) would block the task runtime (in this case `asyncio`), so no other tasks could execute until this call would finish.

PR #225, opened 2025-09-05
Verify pre-commit hooks pass in PR

robertschonfeld:
I decided it's best to only depend on the pre-commit pip package rather than:
- https://pre-commit.ci/ - this is a service recommended by the developer, I find it maybe a bit unnecessary, it's free for open source repos but otherwise costs 20$/month for startups
- https://github.com/pre-commit/action is in "maintenance mode" in favor of the service

### DIFF
--- a/.github/workflows/workflow_pr.yml
+++ b/.github/workflows/workflow_pr.yml
@@ -84,6 +84,7 @@ jobs:
 
       - name: Run pre-commit
         run: uvx pre-commit run --all-files --show-diff-on-failure --color always
+
   ready:
     name: Ready
     needs: [all_sdks_same_version, pre_commit]
@@ -91,4 +92,3 @@ jobs:
     steps:
       - name: Ready
         run: echo "Ready"
-


.github/workflows/workflow_pr.yml
```diff
@@ -70,9 +70,30 @@ jobs:
         run: |
           bun run utils/bump-release-version.ts check
 
+  pre_commit:
+    name: Verify pre-commit hooks pass
+    runs-on: public-amd64
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'ready') }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
```
jerguslejko, line 79: you can make this much faster and less code by using `uv`

```yaml
- name: Setup uv
  uses: astral-sh/setup-uv@v6

- name: Run pre-commit
  uses: uvx pre-commit run --all-files --show-diff-on-failure --color always
```

caching should work ootb

PR #229, opened 2025-09-07
[proto] add `Struct` type: vol #1

jerguslejko:
* add new `Struct` value type, equivalent to `HashMap<String, Value>` (recursive)
* add proto helpers (also few missing for `Value`)

I will add tests in another PR (need this to be merged upstream)

### DIFF
--- a/topk-js/index.d.ts
+++ b/topk-js/index.d.ts
@@ -63,11 +63,6 @@ export interface CollectionFieldSpec {
 export type ConsistencyLevel =  'indexed'|
 'strong';
 
-export interface CreateCollectionOptions {
-  name: string
-  schema: Record<string, schema.FieldSpec>
-}
-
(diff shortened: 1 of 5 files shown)


topk-rs/src/proto/data/v1/data_ext/value.rs
```diff
@@ -153,6 +163,22 @@ impl Value {
         }
     }
 
+    /// Create a struct value from a map of values.
+    pub fn r#struct(values: HashMap<impl Into<String>, Value>) -> Self {
```
marekgalovic, line 167:
```rust
pub fn r#struct<K: Into<String>>(values: impl IntoIterator<Item = (K, Value)>) -> Self {...}
```
is more ergonomic to use.

marekgalovic (approved):

PR #233, opened 2025-09-09
feat: Python AsyncClient

petertoth:
Based off #224 

* adds `AsyncClient` which implements async collection(`get(), count(), query(), upsert(), delete()`) and async collections(`get(), list(), create(), delete()`) clients

```py
import asyncio

from topk_sdk import AsyncClient

client = AsyncClient(
    api_key="YOUR_API_KEY",
    region="aws-us-east-1-elastica"
)

async def main():
    # Create a collection
    await client.collections().create("books", {
        "title": text().required().index(semantic_index())
    })

    # Upsert documents
    lsn = await client.collection("books").upsert([
        {"_id": "1", "title": "The Great Gatsby", "rating": 5},
        {"_id": "2", "title": "1984", "rating": 4}
    ])

    # Query documents
    results = await client.collection("books").query(
        select("title", "rating").topk(field("rating"), 10)
    )

    print(results)

# Run the async function
asyncio.run(main())
```

### DIFF
--- a/topk-py/src/client/async/collection.rs
+++ b/topk-py/src/client/async/collection.rs
@@ -1,4 +1,5 @@
-use crate::data::value::{NativeValue, Value};
+use crate::client::Document;
+use crate::data::value::Value;
 use crate::error::RustError;
 use crate::query::{ConsistencyLevel, Query};
 use pyo3::prelude::*;
@@ -38,10 +39,12 @@ impl AsyncCollectionClient {
                 .await
                 .map_err(RustError)?;
 
-            Ok(docs
+            let docs: HashMap<String, Document> = docs
                 .into_iter()
-                .map(|(id, doc)| (id, doc.into_iter().map(|(k, v)| (k, v.into())).collect()))
-                .collect::<HashMap<String, HashMap<String, NativeValue>>>())
+                .map(|(id, doc)| (id, Document::from(doc)))
+                .collect();
+
+            Ok(docs)
         })
         .map(|result| result.into())
     }
@@ -86,10 +89,9 @@ impl AsyncCollectionClient {
                 .await
                 .map_err(RustError)?;
 
-            Ok(docs
-                .into_iter()
-                .map(|d| d.fields.into_iter().map(|(k, v)| (k, v.into())).collect())
-                .collect::<Vec<HashMap<String, NativeValue>>>())
+            let docs: Vec<Document> = docs.into_iter().map(|d| d.into()).collect();
+
+            Ok(docs)
         })
         .map(|result| result.into())
     }
--- a/topk-py/tests/test_async_collection.py
+++ b/topk-py/tests/test_async_collection.py
@@ -1,6 +1,5 @@
 import asyncio
 import pytest
-from topk_sdk import data, error, query, schema
 from topk_sdk.query import select, field, literal
 
 from . import AsyncProjectContext
@@ -77,7 +76,6 @@ async def test_async_query(async_ctx: AsyncProjectContext):
     assert results == [{"_id": "1984", "title": "1984", "published_year": 1949, "literal": 1.0}]
 
 
-
 test_documents = [
     {
         "_id": "test_doc",
@@ -87,6 +85,7 @@ test_documents = [
         "summary_embedding": [0.5] * 16,
     }
 ]
+
 
 @pytest.mark.asyncio
 async def test_async_collection_parallel_queries(async_ctx: AsyncProjectContext):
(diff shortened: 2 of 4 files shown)


docs/concepts/async-await.mdx
```diff
@@ -0,0 +1,203 @@
+---
+title: "Async/Await"
+---
+
+TopK SDK provides full async/await support for the Python SDK as well as the Javascript SDK.
+
+## Python SDK
+
+`AsyncClient` is an asynchronous client that exposes a coroutine-based API, enabling non-blocking interaction with TopK through `asyncio`:
```
jerguslejko, line 9: `asyncio` is just one of python runtimes (like `tokio`), there are others. 

we can say something simpler like `AsyncClient` is an async version of the client that allows to use `async`/`await` keywords in python

topk-py/src/client/async/collection.rs
```diff
@@ -0,0 +1,139 @@
...
+        let collection = self.collection.clone();
+
+        future_into_py(py, async move {
+            let docs = client
+                .collection(&collection)
+                .get(ids, fields, lsn, consistency.map(|c| c.into()))
+                .await
+                .map_err(RustError)?;
+
+            Ok(docs
```
jerguslejko, line 41: nit: having `Document` here would be nicer

topk-py/src/client/async/collection.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub fn get(
+        &self,
+        py: Python<'_>,
+        ids: Vec<String>,
+        fields: Option<Vec<String>>,
+        lsn: Option<String>,
+        consistency: Option<ConsistencyLevel>,
+    ) -> PyResult<PyObject> {
+        let client = self.client.clone();
+        let collection = self.collection.clone();
```
jerguslejko, line 32: nit: could we change `collection` from `String` to `Arc<String>` so cloning here is cheaper?

topk-py/src/client/async/collections.rs
```diff
@@ -0,0 +1,95 @@
...
+        schema: HashMap<String, FieldSpec>,
+    ) -> PyResult<PyObject> {
+        let client = self.client.clone();
+
+        future_into_py(py, async move {
+            let collection = client
+                .collections()
+                .create(
+                    &collection_name,
+                    schema
```
jerguslejko, line 66: nit: having `Schema` object here would be nicer so we don't have to duplicate the pb conversion

topk-py/src/client/async/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[new]
+    #[pyo3(signature = (api_key, region, host="topk.io".into(), https=true, retry_config=None))]
+    pub fn new(
+        api_key: String,
+        region: String,
+        host: String,
+        https: bool,
+        retry_config: Option<RetryConfig>,
+    ) -> Self {
+        let client = Arc::new(topk_rs::Client::new({
```
jerguslejko, line 30: nit: move client init code to a helper in `topk-py/src/client/mod.rs`

topk-py/tests/test_async_collection.py
```diff
@@ -0,0 +1,592 @@
+import asyncio
```
jerguslejko, line 1: the sync and async tests are not testing the same logic twice. for example `test_async_upsert_to_non_existent_collection` has nothing to do with sync/async.

I would just keep minimum number of tests in the async version, such that we test the code is piped properly, but no need to test for behaviour

topk-py/src/client/async/collection.rs
petertoth: added here https://github.com/topk-io/topk/pull/233/commits/fa0ace7ef1736c37ed3ec3cc1f16be5c6415b08a

docs/concepts/async-await.mdx
petertoth: updated https://github.com/topk-io/topk/pull/233/commits/71d9a14ae1ddea5990f22d0a4a32745e511bdd33

topk-py/src/client/async/collection.rs
petertoth: updated in https://github.com/topk-io/topk/pull/233/commits/71d9a14ae1ddea5990f22d0a4a32745e511bdd33

topk-py/src/client/async/collections.rs
petertoth: updated in https://github.com/topk-io/topk/pull/233/commits/71d9a14ae1ddea5990f22d0a4a32745e511bdd33

jerguslejko (approved):

PR #245, opened 2025-09-19
[WIP] int8 vector

robertschonfeld:


topk-rs/src/proto/data/v1/data_ext/mod.rs
```diff
@@ -55,3 +64,31 @@ impl IntoListValues for Vec<String> {
         list::Values::String(list::String { values: self })
     }
 }
+
+fn transmute_vec_i8_to_u8(vec: Vec<i8>) -> Vec<u8> {
```
marekgalovic, line 68: Since we're transmuting into the same container into a different type with the same size & we can use
```rust
fn transmute_vec_i8_to_u8(x: Vec<i8>) -> Vec<u8> {
    unsafe {
        std::mem::transmute(x)
    }
}
```
There is also https://docs.rs/bytemuck/latest/bytemuck/allocation/fn.cast_vec.html which we could use in public APIs.

PR #415, opened 2026-04-03
[cli] topk-cli

petertoth:


topk-cli/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+
+Alternatively, pass `--api-key` as a flag on any command.
+
+## Commands
+
+### `ask` — Get a grounded answer with citations from documents
+
+```bash
+topk ask "my question"
```
jerguslejko: needs a `--dataset` flag

topk-cli/README.md
```diff
@@ -0,0 +1,189 @@
...
+| ----------- | -------- | ----------------------------------------------------------------------------------- |
+| `--dataset` | Yes      | Dataset to search (repeatable, e.g. `-d ds1 -d ds2`)                               |
+| `--mode`    | No       | Response mode: `auto` (default), `summarize`, `research`                            |
+| `--field`   | No       | Metadata field to include in results (repeatable, e.g. `-f title -f author`)        |
+
+
+The query can also be piped via stdin:
+
+```bash
+echo "my question" | topk ask
```
jerguslejko: same here

topk-cli/README.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+---
+
+### `search` — Find relevant passages in documents
+
+```bash
+topk search "my query"
```
jerguslejko: also needs dataset

topk-cli/README.md
```diff
@@ -0,0 +1,189 @@
...
+| Variable       | Required                | Description                                       |
+| -------------- | ----------------------- | ------------------------------------------------- |
+| `TOPK_API_KEY` | Yes or pass `--api-key` | Your API key. [Get your API key](https://console.topk.io) |
+
+
+Alternatively, pass `--api-key` as a flag on any command.
+
+## Commands
+
+### `ask` — Get a grounded answer with citations from documents
```
jerguslejko: nit, drop the `..from documents`?

topk-cli/README.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+| Argument  | Required | Description  |
+| --------- | -------- | ------------ |
+| `DATASET` | Yes      | Dataset name |
+
+#### Create a dataset:
+
+```bash
+topk dataset create --region us-east-1 my-dataset
```
jerguslejko: nit: use a proper region name

topk-cli/src/lib.rs
```diff
@@ -0,0 +1,94 @@
+pub mod cache;
+pub mod client;
+pub mod commands;
+pub mod config;
+pub mod datasets;
+pub mod output;
+pub mod util;
+
+#[cfg(test)]
+pub mod test_context {
```
jerguslejko: move: move to separate module

topk-cli/src/main.rs
```diff
@@ -0,0 +1,336 @@
...
+use topk::datasets::{ensure_unique_region, get_region, make_cached_datasets_client};
+use topk::output::{Output, OutputFormat};
+use topk::util::resolve_query;
+
+use anyhow::Result;
+use clap::{CommandFactory, Parser, Subcommand};
+use clap_complete::{generate, Shell};
+use topk_rs::proto::v1::ctx::doc::DocId;
+
+use commands::{ask, auth, dataset, delete, list, search, upload};
```
jerguslejko: nit: we usually order imports in three groups `{std}\n{third party}\n{crate}`

topk-cli/src/main.rs
```diff
@@ -0,0 +1,336 @@
...
+        #[arg(short = 'd', long = "dataset")]
+        datasets: Vec<String>,
+        /// Number of results to return
+        #[arg(short = 'k', long, default_value = "10")]
+        top_k: u32,
+        /// Metadata fields to include in results (repeatable)
+        #[arg(short = 'f', long = "field")]
+        fields: Option<Vec<String>>,
+        /// Save search results content (images, text chunks) to a directory
+        #[arg(long, value_name = "DIR", num_args = 0..=1, default_missing_value = ".")]
```
jerguslejko: what does `num_args = 0..=1` do? we don't use it with other falgs

topk-cli/src/main.rs
```diff
@@ -0,0 +1,336 @@
...
+        action: dataset::DatasetAction,
+    },
+
+    /// Generate shell completion script
+    #[command(hide = true)]
+    Completions { shell: Shell },
+}
+
+#[tokio::main]
+async fn main() -> std::process::ExitCode {
```
jerguslejko: nit: import `ExitCode` on top

topk-cli/src/main.rs
```diff
@@ -0,0 +1,336 @@
...
+    if let Err(e) = run(cli, &output).await {
+        output.error(&e);
+        return std::process::ExitCode::FAILURE;
+    }
+
+    std::process::ExitCode::SUCCESS
+}
+
+async fn run(cli: Cli, output: &Output) -> Result<()> {
+    let command = cli.command.expect("checked above");
```
jerguslejko: instead of `.expect()` here you can `match` above to make this state impossible

topk-cli/src/main.rs
```diff
@@ -0,0 +1,336 @@
...
+            let Some(api_key) = auth::resolve(cli.api_key, &host, https, false)? else {
+                return Ok(());
+            };
+
+            let mut client =
+                make_cached_datasets_client(make_global_client(&api_key, &host, https));
+
+            match action {
+                dataset::DatasetAction::List => {
+                    let result = dataset::list(&mut client).await?;
```
jerguslejko: you can pass owned client instead of `&mut client` since only one cmd can run at the same time

topk-cli/src/main.rs
```diff
@@ -0,0 +1,336 @@
...
+        Commands::Upload {
+            dataset,
+            recursive,
+            concurrency,
+            yes,
+            dry_run,
+            wait,
+            timeout,
+            pattern,
+        } => {
```
jerguslejko: instead of enumerating all properties here and then below (`upload::run`), you can define the cmd as `Commands::Upload(UploadArgs)` and then you can just match on `Commands::Upload(args)` and pass that directly. 

should cleanup the code a lot

topk-cli/src/main.rs
```diff
@@ -0,0 +1,336 @@
...
+                &upload::run(
+                    &client,
+                    &dataset,
+                    &pattern,
+                    recursive,
+                    concurrency as usize,
+                    yes,
+                    dry_run,
+                    wait,
+                    Some(std::time::Duration::from_secs(timeout)),
```
jerguslejko: you can define `timeout` as `Duration` directly on the struct

topk-cli/src/main.rs
```diff
@@ -0,0 +1,362 @@
...
+                    &client,
+                    query,
+                    datasets,
+                    mode.map(|m| m.into()),
+                    fields,
+                    output,
+                )
+                .await?,
+            )?;
+        }
```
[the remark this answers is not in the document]
jerguslejko: @petertoth it shouldn't default to all

topk-cli/src/output.rs
```diff
@@ -0,0 +1,166 @@
...
+pub const RESET: &str = "\x1b[0m";
+pub const DIM: &str = "\x1b[2m";
+pub const BOLD: &str = "\x1b[1m";
+pub const BLUE: &str = "\x1b[34m";
+
+#[derive(Debug, Clone, Copy, clap::ValueEnum, Default)]
+pub enum OutputFormat {
+    #[default]
+    #[value(alias = "text")]
+    HumanReadable,
```
jerguslejko: nit: `Text`?

topk-cli/src/output.rs
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    pub fn is_human(&self) -> bool {
+        matches!(self.format, OutputFormat::HumanReadable)
+    }
+
+    pub fn confirm(&self, prompt: &str) -> std::io::Result<bool> {
+        if matches!(self.format, OutputFormat::HumanReadable)
+            && std::io::stdin().is_terminal()
```
jerguslejko: why does `stdin` have to be terminal?

topk-cli/src/output.rs
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    pub fn error(&self, e: &anyhow::Error) {
+        match self.format {
+            OutputFormat::Json => {
+                let payload = serde_json::json!({ "error": format!("{:#}", e) });
+                eprintln!(
+                    "{}",
+                    serde_json::to_string(&payload)
```
jerguslejko: doesn't `serde_json::Value` already implement `Display`?

topk-cli/src/cache.rs
```diff
@@ -0,0 +1,206 @@
...
+pub(crate) fn load() -> DatasetIndex {
+    let path = match cache_path() {
+        Some(p) => p,
+        None => return DatasetIndex::default(),
+    };
+    let content = match std::fs::read_to_string(&path) {
+        Ok(c) => c,
+        Err(_) => return DatasetIndex::default(),
+    };
+    toml::from_str(&content).unwrap_or_default()
```
jerguslejko: we should error out/print a helpful error message when the file is corrupted

topk-cli/src/cache.rs
```diff
@@ -0,0 +1,206 @@
...
+        self.datasets.retain(|d| d.name != name);
+    }
+}
+
+pub fn cache_path() -> Option<PathBuf> {
+    dirs::config_dir().map(|d| d.join("topk").join("datasets.toml"))
+}
+
+pub(crate) fn load() -> DatasetIndex {
+    let path = match cache_path() {
```
jerguslejko: `unwrap_or_default()`

topk-cli/src/cache.rs
```diff
@@ -0,0 +1,206 @@
...
+pub struct DatasetEntry {
+    pub name: String,
+    pub region: String,
+    #[serde(default)]
+    pub org_id: String,
+    #[serde(default)]
+    pub project_id: String,
+    #[serde(default)]
+    pub created_at: String,
+}
```
jerguslejko: we don't need to cache other attributes apart from `name -> region` mapping. `name` is guaranteed to be unique

topk-cli/src/cache.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /// Evicts the entry for `name` because the backend reported it stale.
+    ///
+    /// Semantically distinct from `remove` even though the implementation is the same:
+    /// use `bust` on negative feedback (e.g. `NotFound` from a regional call) so the
+    /// next lookup re-fetches from the backend.
+    pub fn bust(&mut self, name: &str) {
+        self.datasets.retain(|d| d.name != name);
+    }
```
jerguslejko: seems like a client-side concern, we can have just one method for this

topk-cli/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+/// Saves the config file, creating parent directories as needed.
+pub fn save(config: &Config) -> Result<()> {
+    let path =
+        config_path().ok_or_else(|| anyhow::anyhow!("could not determine config directory"))?;
+    if let Some(parent) = path.parent() {
+        std::fs::create_dir_all(parent)?;
+    }
+    std::fs::write(&path, toml::to_string_pretty(config)?)?;
+    Ok(())
+}
```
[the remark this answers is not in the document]
jerguslejko: we should restrict to only the user @petertoth

topk-cli/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub api_key: Option<String>,
+}
+
+/// Returns the path to the config file
+pub fn config_path() -> Option<PathBuf> {
+    dirs::config_dir().map(|d| d.join("topk").join("config.toml"))
+}
+
+/// Loads the config file. Returns an empty config on any read or parse error.
+pub fn load() -> Config {
```
jerguslejko, line 19: `load` and `save` are duplicated between config/dataset cache

topk-cli/src/datasets.rs
```diff
@@ -0,0 +1,398 @@
+use async_trait::async_trait;
+use topk_rs::{
+    proto::v1::control::{CreateDatasetResponse, GetDatasetResponse, ListDatasetsResponse},
+    Client, Error,
+};
+
+use crate::cache::{self, DatasetEntry, DatasetIndex};
+
+#[async_trait(?Send)]
+pub trait DatasetsClient {
```
jerguslejko, line 10: IMO we can just have `topk_rs::Client` and a separate `DatasetCache`. no need for a trait with two different impls. we can all the cache from the CLI level

topk-cli/src/commands/ask.rs
```diff
@@ -0,0 +1,288 @@
...
+        .join(" ");
+    format!("{BOLD}Facts:{RESET}\n{facts_text}")
+}
+
+fn render_refs_section(
+    refs: &HashMap<String, SearchResult>,
+    saved: &HashMap<String, PathBuf>,
+) -> String {
+    if refs.is_empty() {
+        return String::new();
```
jerguslejko: `None`

topk-cli/src/commands/dataset.rs
```diff
@@ -0,0 +1,337 @@
...
+
+    fn try_from(resp: GetDatasetResponse) -> Result<Self, Error> {
+        Ok(Self {
+            dataset: resp
+                .dataset
+                .ok_or_else(|| Error::MalformedResponse("dataset missing from get response".to_string()))?
+                .into(),
+        })
+    }
+}
```
jerguslejko, line 137: the "command" code is littered with output types & proto conversions. we can have a custom _client_ on top of `topk_rs` and do conversions there

topk-cli/src/commands/dataset.rs
```diff
@@ -0,0 +1,337 @@
...
+
+/// `topk dataset delete`
+pub async fn delete<C: DatasetsClient + ?Sized>(
+    client: &mut C,
+    name: &str,
+    yes: bool,
+    output: &Output,
+) -> Result<DeleteDatasetResult, Error> {
+    if !yes && !output.confirm(&format!("Delete dataset '{}'? ", name))? {
```
jerguslejko: wdyt about passing `yes` las a 2nd arg to `option`?

topk-cli/src/commands/auth.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+        if self.env_var_still_set {
+            lines.push(
+                "Note: TOPK_API_KEY is still set in your environment; unset it to fully log out."
+                    .to_string(),
+            );
+        }
+        lines.join("\n")
+    }
+}
```
jerguslejko: imo we can simplify. no need to put everything in the output, we can use logging for extra info

topk-cli/src/commands/auth.rs
```diff
@@ -0,0 +1,174 @@
...
+            );
+        }
+        eprintln!("No API key found.");
+        return prompt_menu(host, https)?
+            .ok_or_else(|| anyhow::anyhow!("API key not set. Run `topk auth` to configure one"))
+            .map(Some);
+    }
+
+    // Explicit auth: show current status, then always prompt.
+    if let Ok(key) = std::env::var("TOPK_API_KEY") {
```
jerguslejko: we already read the api key via cli args, why read it again?

jerguslejko (changes requested):

topk-cli/src/commands/upload/progress.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/// Build a reporter for the upload phase.
+///
+/// Falls back to a no-op progress bar when the user asked for JSON output or when
+/// stderr is not a TTY (e.g. piping logs to a file). [`Output`] is always stored
+/// so per-file errors can be printed consistently in human mode after the bar finishes.
+pub fn upload_reporter(total: usize, output: &Output) -> Box<dyn ProgressReporter> {
+    let output = *output;
+    if !output.is_human() || !std::io::stderr().is_terminal() {
```
jerguslejko: wdyt about `Output` checking `is_terminal()` internally?

topk-cli/src/commands/upload/progress.rs
```diff
@@ -0,0 +1,125 @@
...
+            total: total as u64,
+            completed: AtomicU64::new(0),
+            failed: AtomicU64::new(0),
+            output,
+        }
+    }
+}
+
+impl ProgressReporter for BarReporter {
+    fn on_upload(&self, ok: bool) {
```
jerguslejko: @petertoth do we retry uploads?

topk-cli/src/commands/upload/plan.rs
```diff
@@ -0,0 +1,138 @@
...
+
+pub enum PlanOutcome {
+    NoFiles { message: String },
+    Files { files: Vec<UploadFile>, totals: Totals },
+}
+
+/// Resolves files matching `pattern` relative to `cwd` and computes totals.
+///
+/// Pure sync — no network, no progress, no interactive I/O — so it's trivial
+/// to unit test the "no files" hint, absolute paths, and recursive behavior.
```
jerguslejko: drop comments like this. they are just artifacts of LLM thinking

topk-cli/src/commands/upload/result.rs
```diff
@@ -0,0 +1,208 @@
...
+pub struct UploadError {
+    pub doc_id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub path: Option<PathBuf>,
+    pub error: String,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(tag = "kind", rename_all = "snake_case")]
+pub enum UploadOutcome {
```
jerguslejko: what the reasoning behind using `UploadOutcome`?

topk-cli/src/commands/upload/mod.rs
```diff
@@ -0,0 +1,391 @@
...
+                "--dry-run",
+            ])
+            .output()
+            .unwrap();
+        assert!(
+            out.status.success(),
+            "{}",
+            String::from_utf8_lossy(&out.stderr)
+        );
+        let result: UploadResult = serde_json::from_slice(&out.stdout).unwrap();
```
jerguslejko: nit: you can add a trait that adds `.json<T>() -> Result` method to make the tests cleaner

topk-cli/src/commands/upload/uploading.rs
```diff
@@ -0,0 +1,86 @@
...
+
+            async move {
+                let result = async {
+                    let input = InputFile::from_path(&file.path)?;
+                    let handle = client
+                        .dataset(&dataset)
+                        .upsert_file(
+                            file.doc_id.clone(),
+                            input,
+                            std::iter::empty::<(String, Value)>(),
```
jerguslejko: why not `HashMap::default()`?

topk-cli/src/commands/upload/result.rs
```diff
@@ -0,0 +1,208 @@
...
+            errors: vec![],
+            processed: Some(true),
+        };
+        let json = serde_json::to_value(&outcome).unwrap();
+        assert_eq!(json["kind"], "uploaded");
+        assert_eq!(json["uploaded"], 2);
+        assert_eq!(json["processed"], true);
+        let parsed: UploadOutcome = serde_json::from_value(json).unwrap();
+        assert!(matches!(parsed, UploadOutcome::Uploaded { .. }));
+    }
```
jerguslejko: what are we actually testing here? JSON serialization?

topk-cli/src/main.rs
petertoth: it allows user to pass `--output-dir` without a value which will then default to `.`

jerguslejko: I would make the default value here required.

topk-cli/src/commands/upload/progress.rs
petertoth: they are retried in topk-rs `upsert_file`:

https://github.com/topk-io/topk/blob/5900f271ed3a8b1728c4586963afdfd7d39729cd/topk-rs/src/client/dataset.rs#L109-L130

topk-cli/README.md
```diff
@@ -14,7 +14,7 @@ brew install topk
 Run:
 
 ```bash
-topk auth
+topk auth login
```
jerguslejko: @petertoth wdyt about promoting these to `topk login` and `topk logout` ?

topk-cli/src/commands/auth.rs
```diff
@@ -0,0 +1,174 @@
...
+                "Note: TOPK_API_KEY is still set in your environment; unset it to fully log out."
+                    .to_string(),
+            );
+        }
+        lines.join("\n")
+    }
+}
+
+pub fn logout() -> Result<LogoutResult> {
+    let path = config::config_path()
```
jerguslejko: we should load/resolve config in `main.rs`

topk-cli/src/commands/dataset.rs
```diff
@@ -0,0 +1,346 @@
...
+    client.create(name, region).await?.try_into()
+}
+
+/// `topk dataset delete`
+pub async fn delete<C: DatasetsClient>(
+    mut client: C,
+    name: &str,
+    yes: bool,
+    output: &Output,
+) -> Result<DeleteDatasetResult, Error> {
```
jerguslejko, line 214: this could just accept `client: C, args: DeleteArgs`

topk-cli/README.md
```diff
@@ -0,0 +1,198 @@
...
+```bash
+topk upload '*.pdf' --dataset my-dataset
+topk upload 'docs/**/*.md' --dataset my-dataset
+topk upload docs --dataset my-dataset -r
+```
+
+
+| Argument    | Required | Description                                                               |
+| ----------- | -------- | ------------------------------------------------------------------------- |
+| `PATTERN`   | Yes      | A file path, directory, or glob pattern |
```
jerguslejko: lets make `PATTERN` repeatable too

topk-cli/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+async fn run(cli: Cli, output: &Output) -> Result<(), Error> {
+    let Cli {
+        command,
+        api_key,
+        host,
+        https,
+        output: _,
+    } = cli;
```
jerguslejko: nit: just use `cli.foo`

topk-cli/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+                cfg.api_key = Some(api_key.clone());
+                config::save(&cfg)?;
+                if output.is_json() {
+                    output.print(&result)?;
+                } else {
+                    output.success("API key saved.");
+                }
+            } else {
+                output.print(&result)?;
+            }
```
jerguslejko: this should exit with an error, right?

topk-cli/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+    } = cli;
+
+    let mut cfg = config::load();
+
+    match command {
+        Some(Commands::Login) => {
+            let result = login::run(&host, https)?;
+            if let Some(api_key) = &result.api_key {
+                cfg.api_key = Some(api_key.clone());
+                config::save(&cfg)?;
```
jerguslejko: nit: `config::set_api_key()`

topk-cli/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+                output.print(&result)?;
+            }
+            Ok(())
+        }
+
+        Some(Commands::Logout) => {
+            let result = logout::run(&cfg);
+            if result.cleared {
+                cfg.api_key = None;
+                config::save(&cfg)?;
```
jerguslejko: `config::clear()` (or `clear_api_key()`)

topk-cli/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        Some(Commands::List(args)) => {
+            let api_key = require_api_key(api_key)?;
+            let mut datasets_client =
+                make_cached_datasets_client(make_global_client(&api_key, &host, https));
+            let region = get_region(&mut datasets_client, &args.dataset).await?;
+            let client = make_client(&api_key, &region, &host, https);
+
+            // JSON mode streams NDJSON progressively inside run(); only render the table in non-json mode.
```
jerguslejko: we should use a stream here

topk-cli/src/util.rs
```diff
@@ -0,0 +1,473 @@
+use std::{
```
jerguslejko: split into submodules

topk-cli/src/util.rs
```diff
@@ -0,0 +1,473 @@
...
+}
+
+impl<'de> Deserialize<'de> for MimeType {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let s = String::deserialize(deserializer)?;
+        Ok(MimeType::from(s.as_str()))
+    }
+}
+
+pub(crate) fn doc_id_from_path(path: &Path) -> Result<String, Error> {
```
jerguslejko: should live in `upload`

topk-cli/src/output.rs
```diff
@@ -0,0 +1,194 @@
...
+pub const BLUE: &str = "\x1b[34m";
+
+#[derive(Debug, Clone, Copy, clap::ValueEnum, Default)]
+pub enum OutputFormat {
+    #[default]
+    Text,
+    Json,
+}
+
+pub trait RenderForHuman: Serialize {
```
jerguslejko: this is basically the `Display` trait, we should use that

topk-cli/src/output.rs
```diff
@@ -0,0 +1,194 @@
...
+                Ok(())
+            }
+        }
+    }
+
+    pub fn print_json_line<T: Serialize>(&self, value: &T) -> Result<(), serde_json::Error> {
+        let stdout = std::io::stdout();
+        let mut lock = stdout.lock();
+        serde_json::to_writer(&mut lock, value)?;
+        writeln!(&mut lock).map_err(serde_json::Error::io)?;
```
jerguslejko: why not just `println!()`?

topk-cli/src/cache.rs
```diff
@@ -0,0 +1,190 @@
+use std::path::PathBuf;
```
jerguslejko: we could merge this cache with the `config` module

topk-cli/src/commands/search.rs
```diff
@@ -0,0 +1,343 @@
...
+                None
+            } else {
+                Some(text.to_string())
+            }
+        }
+        _ => None,
+    }
+}
+
+pub fn format_content_text(content: Option<&Content>) -> String {
```
jerguslejko: should return `Option<String>`

topk-cli/src/commands/ask.rs
```diff
@@ -0,0 +1,290 @@
...
+    #[arg(long, value_name = "DIR")]
+    pub output_dir: Option<PathBuf>,
+}
+
+/// `topk ask`
+pub async fn run(client: &Client, args: &AskArgs, output: &Output) -> Result<AskResult, Error> {
+    let query = resolve_query(args.query.clone())
+        .map_err(|e| Error::Internal(e.to_string()))?
+        .ok_or_else(|| {
+            Error::Internal(
```
jerguslejko: should this be an internal error?

topk-cli/src/commands/ask.rs
```diff
@@ -0,0 +1,290 @@
...
+    fn render(&self) -> impl Into<String> {
+        let facts = render_facts(&self.facts);
+        let refs = render_refs(&self.refs)
+            .map(|r| format!("\n{r}"))
+            .unwrap_or_default();
+        format!("{facts}{refs}")
+    }
+}
+
+fn has_non_text_search_results(
```
jerguslejko: we should store _all_ results, not just non-text

topk-cli/src/commands/list.rs
```diff
@@ -0,0 +1,208 @@
...
+        .dataset(&args.dataset)
+        .list(args.fields.clone(), None)
+        .await?
+        .into_inner();
+
+    let mut entries = Vec::new();
+    while let Some(entry) = stream.next().await {
+        let row: ListEntryRow = entry?.into();
+        // Stream each entry as NDJSON so the caller sees results progressively.
+        // The caller must not call output.print in JSON mode — entries are already written.
```
jerguslejko: why don't we let the caller do the printing?

topk-cli/src/commands/delete.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/// `topk delete`
+pub async fn run(
+    client: &Client,
+    args: &DeleteArgs,
+    output: &Output,
+) -> Result<DeleteResult, Error> {
+    if !output.confirm_or_yes(&format!("Delete document '{}'? ", args.id), args.yes)? {
+        return Ok(DeleteResult {
```
jerguslejko: why not return `Result<Handle, Error>` and have `Error::UserCancelled`?

topk-cli/src/commands/ask.rs
```diff
@@ -0,0 +1,290 @@
...
+            let refs_inline = if fact.ref_ids.is_empty() {
+                None
+            } else {
+                let ids = fact
+                    .ref_ids
+                    .iter()
+                    .map(|id| format!("[{id}]"))
+                    .collect::<Vec<_>>()
+                    .join(" ");
+                Some(format!("{BLUE}{ids}{RESET}"))
```
jerguslejko: we can use `colored` crate instead of raw color codes

topk-cli/src/commands/ask.rs
```diff
@@ -0,0 +1,290 @@
...
+            let parts: Vec<&str> = [fact.fact.as_str()]
+                .into_iter()
+                .chain(refs_inline.as_deref())
+                .filter(|s| !s.is_empty())
+                .collect();
+            if parts.is_empty() {
+                None
+            } else {
+                Some(parts.join(" "))
+            }
```
jerguslejko: why put it in an iterator, if it's only one string?

topk-cli/src/commands/ask.rs
```diff
@@ -0,0 +1,290 @@
...
+        .into_iter()
+        .map(|(k, v)| {
+            let path = if let Some(ref dir) = output_dir {
+                write_result_content(dir, &k, &v).map_err(Error::IoError)?
+            } else {
+                None
+            };
+            Ok::<_, Error>((k, SearchResult { result: v, path }))
+        })
+        .collect::<Result<_, _>>()?;
```
jerguslejko: this should be simplified to just `write_results(answer, dir)?`

topk-cli/src/commands/delete.rs
```diff
@@ -0,0 +1,159 @@
...
+                "doc-to-keep",
+            ])
+            .output()
+            .unwrap();
+        assert!(
+            out.status.success(),
+            "{}",
+            String::from_utf8_lossy(&out.stderr)
+        );
+        let result: DeleteResult = serde_json::from_slice(&out.stdout).unwrap();
```
jerguslejko: nit: use the helper trait

topk-cli/src/commands/login.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    Err(Error::Input(anyhow::anyhow!(
+        "API key not set. Set TOPK_API_KEY environment variable or run: topk login"
+    ))
+    .into())
+}
+
+fn prompt_api_key(host: &str, https: bool) -> Result<LoginResult, Error> {
+    let options = ["Create a new API key", "Use an existing API key", "Skip"];
```
jerguslejko: nit: inline

topk-cli/src/commands/login.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+fn prompt_and_save() -> Result<LoginResult, Error> {
+    let password = Password::new().with_prompt("API key").interact();
+
+    let api_key = match password {
+        Ok(password) => password.trim().to_string(),
+        Err(e) => return Err(Error::Input(anyhow::anyhow!(e.to_string()))),
+    };
```
jerguslejko: why not `.interact()?`

topk-cli/src/commands/login.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    if api_key.is_empty() {
+        return Err(Error::Input(anyhow::anyhow!("no API key provided")));
+    }
+
+    Ok(LoginResult {
+        api_key: Some(api_key),
```
jerguslejko: looks like there isn't a way to have `api_key: None`

jerguslejko (approved):

topk-cli/src/commands/logout.rs
```diff
@@ -0,0 +1,26 @@
...
+        }
+    }
+}
+
+/// `topk logout`
+pub fn run(config: &config::Config) -> LogoutResult {
+    LogoutResult {
+        cleared: config.api_key.is_some(),
+    }
+}
```
jerguslejko: this doesn't actually do anything?

topk-cli/src/commands/dataset.rs
```diff
@@ -0,0 +1,350 @@
...
+
+impl From<&topk_rs::proto::v1::control::Dataset> for Dataset {
+    fn from(dataset: &topk_rs::proto::v1::control::Dataset) -> Self {
+        Self {
+            name: dataset.name.clone(),
+            region: dataset.region.clone(),
+            created_at: dataset.created_at.clone(),
+        }
+    }
+}
```
jerguslejko: there is already `impl From<topk_rs::proto::v1::control::Dataset>` so no need for `impl From<&topk_rs::proto::v1::control::Dataset>`, the caller can clone

topk-cli/src/commands/dataset.rs
```diff
@@ -0,0 +1,350 @@
...
+        dataset: String,
+    },
+    /// Create a dataset
+    Create(CreateDatasetArgs),
+    /// Delete a dataset
+    Delete(DeleteDatasetArgs),
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct Dataset {
```
jerguslejko, line 55: why not just use the proto type?

topk-cli/src/commands/ask.rs
```diff
@@ -0,0 +1,290 @@
...
+        )
+        .await?
+        .into_inner();
+
+    let mut answer: Option<Answer> = None;
+    while let Some(item) = stream.next().await {
+        let item = item?;
+        match item.message {
+            Some(ask_result::Message::Reason(r)) => {
+                spinner.println(format!("[thinking] {}", r.thought));
```
jerguslejko: supernit: I think this is `print` not `println` since it doesn't print a new line

topk-cli/src/commands/ask.rs
```diff
@@ -0,0 +1,290 @@
...
+            Mode::Research => topk_rs::proto::v1::ctx::Mode::Research,
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct AskResult {
+    pub(crate) facts: Vec<Fact>,
+    pub(crate) refs: HashMap<String, SearchResult>,
+}
```
jerguslejko, line 43: this is pretty much the `Answer` proto, we could use that directly

topk-cli/src/commands/dataset.rs
```diff
@@ -0,0 +1,350 @@
...
+        let out = cmd()
+            .args(["-o", "json", "dataset", "list"])
+            .output()
+            .unwrap();
+        assert!(
+            out.status.success(),
+            "{}",
+            String::from_utf8_lossy(&out.stderr)
+        );
+        let result: ListDatasetsResult = serde_json::from_slice(&out.stdout).unwrap();
```
jerguslejko: use the `.json()`...

topk-cli/src/util.rs
```diff
@@ -0,0 +1,473 @@
...
+        serializer.serialize_str(&self.to_string())
+    }
+}
+
+impl<'de> Deserialize<'de> for MimeType {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let s = String::deserialize(deserializer)?;
+        Ok(MimeType::from(s.as_str()))
+    }
+}
```
jerguslejko: you can just put `Serialize`/`Deserialize` on the struct directly

topk-cli/src/util.rs
```diff
@@ -0,0 +1,473 @@
...
+    ))
+}
+
+pub fn format_timestamp(rfc3339: &str) -> Option<String> {
+    let dt = rfc3339.parse::<DateTime<Utc>>().ok()?;
+    Some(
+        dt.with_timezone(&Local)
+            .format("%b %-d, %Y %H:%M")
+            .to_string(),
+    )
```
jerguslejko: nit: `f().ok().and_then(|dt| dt.foo())`

topk-cli/src/util.rs
```diff
@@ -0,0 +1,473 @@
...
+        pattern
+            .file_name()
+            .map(|name| name.to_string_lossy().to_string())
+            .unwrap_or_default()
+    } else {
+        matching_components.join(std::path::MAIN_SEPARATOR_STR)
+    };
+
+    (root, matching_pattern)
+}
```
jerguslejko: what's the issue with using `glob::glob` (or `globset`)?

topk-cli/src/util.rs
```diff
@@ -0,0 +1,473 @@
...
+    if path.is_dir() {
+        return collect_directory_files(path, recursive);
+    }
+    let (root, matching_pattern) = split_glob_root(cwd, path);
+    collect_matching_files(&root, &matching_pattern)
+}
+
+pub(crate) fn collect_file(path: &Path) -> Result<UploadFile, Error> {
+    let doc_id = doc_id_from_path(path)?;
+    let size = path.metadata().map(|m| m.len()).unwrap_or(0);
```
jerguslejko: should return error IMO

topk-cli/src/commands/ask.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct AskResult {
+    pub(crate) facts: Vec<Fact>,
+    pub refs: HashMap<String, SearchResult>,
+}
+
+impl From<Answer> for AskResult {
+    fn from(a: Answer) -> Self {
+        Self {
+            facts: a.facts,
+            refs: a.refs.into_iter().collect(),
```
jerguslejko, line 49: not needer?

topk-cli/src/commands/ask.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+}
+
+impl fmt::Display for AskResult {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&render_facts(&self.facts))
+    }
+}
+
+fn render_facts(facts: &[Fact]) -> String {
```
jerguslejko: nit: inline?

topk-cli/src/commands/dataset.rs
```diff
@@ -0,0 +1,343 @@
...
+}
+
+/// `topk dataset list`
+pub async fn list<C: DatasetsClient>(mut client: C) -> Result<ListDatasetsResult, Error> {
+    Ok(client.list().await?.into())
+}
+
+/// `topk dataset get`
+pub async fn get<C: DatasetsClient>(mut client: C, name: &str) -> Result<GetDatasetResult, Error> {
+    client.get(name).await?.try_into()
```
jerguslejko: nit: `Ok(client.get(name).await?.into())`

topk-cli/src/commands/delete.rs
```diff
@@ -0,0 +1,155 @@
...
+    #[arg(short = 'y', long)]
+    pub yes: bool,
+}
+
+/// `topk delete`
+pub async fn run(
+    client: &Client,
+    args: &DeleteArgs,
+    output: &Output,
+) -> Result<DeleteResult, Error> {
```
jerguslejko, line 41: applicable generally: we can consume `Client`/`Args` (maybe also `Output`?) in every cmd, so we can avoid `args.id.clone()`

topk-cli/src/commands/list.rs
```diff
@@ -0,0 +1,221 @@
...
+#[derive(Serialize, Deserialize)]
+pub struct ListResult {
+    pub entries: Vec<ListEntry>,
+}
+
+fn format_status(status: &str) -> (&str, Option<Color>) {
+    match status {
+        "pending" => ("Pending", Some(Color::Yellow)),
+        "ready" => ("Ready", Some(Color::Green)),
+        _ => (status, None),
```
jerguslejko: we also have `error`, in which case I'd display `Error ({status_reason})`

topk-cli/src/commands/list.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let rows = self
+            .entries
+            .iter()
+            .map(|entry| {
+                (
+                    entry.name.clone(),
+                    entry.id.clone(),
+                    format_status(&entry.status),
+                    ByteSize(entry.size).to_string(),
```
jerguslejko: nit: do `str -> ByteSize` in the proto conversion

topk-cli/src/commands/search.rs
```diff
@@ -0,0 +1,267 @@
...
+
+pub fn render_search_result(
+    ref_id: &str,
+    result: &SearchResult,
+    path: Option<&Path>,
+    max_text_len: Option<usize>,
+) -> String {
+    let text = format_reference_detail(result.content.as_ref()).map(|t| match max_text_len {
+        Some(max) if t.chars().count() > max => {
+            format!("{}…", t.chars().take(max).collect::<String>())
```
jerguslejko: we need to render entire search result

topk-cli/src/commands/search.rs
```diff
@@ -0,0 +1,267 @@
...
+        "{} {}, {}, {}",
+        format!("[{ref_id}]").blue(),
+        result.dataset,
+        result.doc_id,
+        result.doc_type,
+    )]
+    .into_iter()
+    .chain(detail)
+    .collect::<Vec<_>>()
+    .join("\n")
```
jerguslejko: why iter when only one item?

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,772 @@
...
+
+impl fmt::Display for UploadResult {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.total == 0 {
+            return f.write_str("No files found for upload.");
+        }
+
+        if self.uploaded == 0 {
+            return f.write_str("Upload skipped.");
+        }
```
jerguslejko, line 67: this will also trigger when all upload fail, right?

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,772 @@
...
+    pub patterns: Vec<String>,
+}
+
+fn make_upload_plan(
+    cwd: &Path,
+    patterns: &[String],
+    recursive: bool,
+) -> Result<Vec<UploadFile>, Error> {
+    let mut files = Vec::new();
+    let mut seen = std::collections::HashSet::new();
```
jerguslejko: nit: import on top

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,772 @@
...
+            if seen.insert(file.doc_id.clone()) {
+                files.push(file);
+            }
+        }
+    }
+
+    Ok(files)
+}
+
+pub(crate) fn doc_id_from_path(path: &Path) -> Result<String, Error> {
```
jerguslejko, line 130: @petertoth wdyt we do a shorter hash here and also append sanitized file name? so something like `my-report-123123123.pdf`?

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,772 @@
...
+fn progress_summary(total: usize, uploaded: usize, failed: usize) -> String {
+    let files = plural(total, "file", "files");
+    match failed {
+        0 => format!("{uploaded}/{total} {files} uploaded"),
+        _ => format!("{uploaded}/{total} {files} uploaded ({failed} failed)"),
+    }
+}
+
+struct NoopReporter {
+    output: Output,
```
jerguslejko: why does `NoopReporter` need `Output`?

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,772 @@
...
+impl BarReporter {
+    fn new(total: usize, output: Output) -> Self {
+        let progress_bar = ProgressBar::new(total as u64);
+        let style =
+            ProgressStyle::with_template("[{elapsed_precise}] [{wide_bar:.cyan/blue}] {msg}")
+                .map(|s| s.progress_chars("=>-"))
+                .unwrap_or_else(|_| ProgressStyle::default_bar());
+        progress_bar.set_style(style);
+        progress_bar.set_message(format!("0/{total} uploaded"));
+        progress_bar.enable_steady_tick(std::time::Duration::from_millis(100));
```
jerguslejko: nit; import on top

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,772 @@
...
+
+        self.progress_bar.set_position(completed);
+        self.progress_bar.set_message(msg);
+    }
+
+    fn finish(self: Box<Self>, total: usize, files: &[FileUpload]) {
+        let uploaded = files.iter().filter(|f| f.result.is_ok()).count();
+        let failed = files.iter().filter(|f| f.result.is_err()).count();
+        self.progress_bar
+            .finish_with_message(progress_summary(total, uploaded, failed));
```
jerguslejko: nit: just pass `files` and derive counts internally?

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,772 @@
...
+        process_fut.await?;
+        true
+    };
+
+    spinner.finish();
+
+    Ok(processed)
+}
+
+fn cancel_on_enter() -> tokio::sync::oneshot::Receiver<()> {
```
jerguslejko: @petertoth wdyt about going "more standard" here and just telling user for `Ctrl+c`? Using "enter" is little weird imo, I often press enter like crazy when I'm waiting for something to finish only to see if it's lagging or not (which in this case cancels the cmd..)

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,772 @@
...
+    }
+
+    if args.dry_run {
+        return Ok(UploadResult {
+            total,
+            uploaded: 0,
+            errors: vec![],
+            processed: None,
+        });
+    }
```
jerguslejko: nit: merge

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,772 @@
...
+
+    let reporter = upload_reporter(total, output);
+    let upload_fut = upload_all(
+        client,
+        &args.dataset,
+        files,
+        args.concurrency as usize,
+        &*reporter,
```
jerguslejko: why `&*`?

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,772 @@
...
+    );
+    let file_uploads = match tokio::time::timeout(args.timeout, upload_fut).await {
```
jerguslejko: @petertoth do we need explicit timeout here? I think we should be able to rely on topk-rs directly

topk-cli/src/util/files.rs
```diff
@@ -0,0 +1,260 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct UploadFile {
+    pub(crate) path: PathBuf,
+    pub(crate) doc_id: String,
+    pub(crate) size: u64,
+    pub(crate) mime_type: MimeType,
+}
+
+pub(crate) fn expand_path(pattern: &str) -> Result<PathBuf, Error> {
+    let expanded = shellexpand::full(pattern)
```
jerguslejko: `::full()` is dangerous, it also expands env vars.. I think there's `::tilde()` or maybe other less powerful ones

topk-cli/src/util/files.rs
```diff
@@ -0,0 +1,260 @@
...
+    pattern: &str,
+    recursive: bool,
+) -> Result<Vec<UploadFile>, Error> {
+    let expanded = expand_path(pattern)?;
+
+    let path = if expanded.is_absolute() {
+        expanded
+    } else {
+        cwd.join(expanded)
+    };
```
jerguslejko: nit: do inside `expand_path`?

topk-cli/src/util/files.rs
```diff
@@ -0,0 +1,260 @@
...
+    pub(crate) mime_type: MimeType,
+}
+
+pub(crate) fn expand_path(pattern: &str) -> Result<PathBuf, Error> {
+    let expanded = shellexpand::full(pattern)
+        .map_err(|err| Error::InvalidArgument(format!("invalid pattern '{}': {}", pattern, err)))?;
+    Ok(PathBuf::from(expanded.as_ref()))
+}
+
+fn is_glob_pattern(pattern: &str) -> bool {
```
jerguslejko, line 34: not sure why this is matching on `?`/`[`?

topk-cli/src/util/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub fn parse_seconds(value: &str) -> Result<Duration, String> {
+    value
+        .parse::<u64>()
+        .map(Duration::from_secs)
+        .map_err(|err| err.to_string())
+}
+
+/// Resolves a query string from an optional CLI argument, falling back to stdin
+pub fn resolve_query(arg: Option<String>) -> Result<String, Error> {
```
jerguslejko: imo we should check `Option<>` on the caller, and only call this fn when `None`.

topk-cli/src/commands/upload.rs
```diff
@@ -0,0 +1,759 @@
...
+        self.progress_bar
+            .finish_with_message(progress_summary(total, files));
+
+        for f in files {
+            if let Err(e) = &f.result {
+                self.output.warn(&format!("  {}: {e}", f.path.display()));
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
petertoth: that's ok for the it's in the stderr

PR #423, opened 2026-04-13
[docs] ctx

petertoth:


README.md
```diff
@@ -5,19 +5,24 @@
 
 # TopK
 
-[TopK](https://topk.io) is a high-performance search engine designed to power modern product experiences. It supports hybrid search out of the box—combining keyword relevance, semantic understanding, and structured filtering in a single query. With a simple and expressive query language, TopK makes it easy to build fast, flexible, and accurate search experiences without stitching together multiple tools.
-
-TopK's query language is designed to be simple yet expressive, allowing you to search using semantic similarity, keyword matching, and filters—all in a single query. SDKs are available for Python and JavaScript.
+TopK is an end-to-end context engine that turns unstructured documents into reliable context for agents and humans.
```
jerguslejko: wdyt about saying `Topk is an end-to-end solution that turns ...`

since the user is interacting with ctx engine for the first time

README.md
```diff
@@ -5,19 +5,24 @@
...
 
-- [Python SDK](./topk-py/README.md) - Python client for TopK
-- [Node.js SDK](./topk-js/README.md) - JavaScript/TypeScript client for Node.js environments
+- [Python SDK](./topk-py/) - Python SDK for TopK API
+- [JavaScript SDK](./topk-js/) - Javascript SDK for TopK API with full Typescript support
+- [Rust SDK](./topk-rs/) - Rust SDK for TopK API
+
+## CLI
+
+- [TopK CLI](./topk-cli/) - CLI for TopK API
```
jerguslejko, line 19: also mention MCP here?

utils/sync-readmes.ts
```diff
@@ -0,0 +1,122 @@
+// usage: bun run utils/sync-readmes.ts
```
jerguslejko, line 1: nice 👍

topk-rs/README.md
```diff
@@ -0,0 +1,146 @@
+# TopK Rust SDK
+
+The TopK Rust library provides convenient access to the TopK API from Rust applications. It is built for asynchronous usage with Tokio and supports document ingestion, grounded Q&A, search, and retrieval workflows.
```
jerguslejko: `and retrieval workflows.` is super vague, I'd just drop iot

topk-py/README.md
```diff
@@ -1,100 +1,154 @@
...
-  "books",
-  schema={
-    "title": text().required().index(semantic_index()),  # Semantic search enabled on title
-  },
+# Ask a question across the dataset
+answer = client.ask(
+    "What are the key findings?",
+    datasets=["my-docs"],
 )
+print(answer.facts)
```
jerguslejko, line 57: @petertoth wdyt we always include response preview? at least the shape of the data

docs/quickstart.mdx
```diff
@@ -1,180 +1,200 @@
...
+      <Step title="Get your API key">
+        ```bash
+        export TOPK_API_KEY="your-api-key"
+        ```
+      </Step>
+      <Step title="Upload a file">
+        Upload a file to your dataset:
+
+        ```bash
+        topk upload ./report.pdf -d quickstart --region aws-us-east-1-elastica
```
jerguslejko: imo we should use the long form `--dataset` (here, and in most other examples) of args since the users are not familiar with the CLI yet

docs/quickstart.mdx
```diff
@@ -1,180 +1,200 @@
...
+        ```
+
+        The `--region` determines where your data is stored. See [available regions](/regions).
```
jerguslejko: @petertoth is this required because we want to automatically create a dataset? It's little weird having this flag on `upload`

docs/cli.mdx
```diff
@@ -0,0 +1,163 @@
...
+```bash
+topk ask "What are the key findings?" -d my-docs
+```
+
+| Flag | Description |
+|------|-------------|
+| `--dataset` / `-d` | Dataset to search (repeatable, at least one required) |
+| `--mode` | Response mode: `auto` (default), `summarize`, or `research` |
+| `--field` / `-f` | Metadata field to include in results (repeatable) |
+| `--output-dir` | Save non-text references to a directory |
```
jerguslejko, line 48: this should be _all_ references, not just non-text. basically this flag is like `--export`. we should store also the response as json there

docs/cli.mdx
```diff
@@ -0,0 +1,163 @@
...
+
+Delete a document from a dataset.
+
+```bash
+topk delete --document-id doc-001 --dataset my-docs
+```
+
+| Flag | Description |
+|------|-------------|
+| `--document-id` | Document ID |
```
jerguslejko: should be just `--id`

docs/core/agent-search.mdx
```diff
@@ -0,0 +1,356 @@
...
+- **facts** - the answer broken into grounded factual statements, with citation identifiers attached to each fact
+- **citations** - references to the underlying source documents and passages
+
+That structure is what makes Agent Search valuable:
+
+- humans can verify where a statement came from
+- agents can decide whether the evidence is strong enough to act on
+- applications can render answers and supporting evidence together
+
+The important property is that Agent Search returns **facts with citations**, not unsupported free-form generation.
```
jerguslejko: I don't get what `not unsupported free-form generation` means

docs/core/agent-search.mdx
```diff
@@ -0,0 +1,356 @@
...
+
+For example, if the query is:
+
+```text
+What does the policy say about contractor access?
```
jerguslejko: @petertoth questions going into the ctx engine should be specific (ie. generated by the agent). `the policy` is ambiguous and will match all policies in the system

docs/core/agent-search.mdx
```diff
@@ -0,0 +1,356 @@
...
+```
+
+you can think of the final Agent Search result like this:
```
jerguslejko: `and the answer looks like:`?

docs/core/agent-search.mdx
```diff
@@ -0,0 +1,356 @@
+---
+title: "Agent Search"
```
jerguslejko: @petertoth any reason not to call this `Ask`?

docs/core/ingest.mdx
```diff
@@ -0,0 +1,322 @@
...
+category, or any custom metadata fields you want to attach to the document.
+
+## Processing documents
+
+After a document is uploaded, TopK returns a **processing handle** — a reference to the processing job.
+The document is not ready for retrieval until it has been processed.
+
+Processing happens asynchronously in the background and may take a few seconds to a few minutes, depending on document size and complexity.
+
+<Info>
```
jerguslejko, line 188: I'd drop this. the paragraph above already mentions that it's asynchronous and needs to be processed. 

this extra comment makes it look kinda slow?

docs/quickstart.mdx
petertoth: yeah ... let me get rid of the region flag and figure out how to prompt user for region if dataset does not exist yet

docs/cli.mdx
petertoth: btw this page will be generated from `topk-cli/README.md` so this one might have old info

docs/core/agent-search.mdx
petertoth: I find `Ask` a bit short and vague. Also it maps 1-1 to the SDK/CLI api which is also used by research so it might be confusing. I was thinking of `Agentic Search` or something similar

jerguslejko: imo `Agentic Search` is too vague and can mean ask/research/etc. Calling it `Ask` sticks to what it does. having `Research` on a separate page is fine

jerguslejko (approved):

PR #434, opened 2026-04-28
[topk-rs] check the file can be opened before upserting

matospiso:


topk-cli/src/util/files.rs
```diff
@@ -61,6 +71,8 @@ pub(crate) fn resolve_files(
 }
 
 pub(crate) fn collect_file(path: &Path) -> Result<UploadFile, Error> {
+    std::fs::File::open(path).map_err(|e| file_access_error(path, e))?;
```
petertoth: you can do just `std::fs::File::open(path)?;` here as we have `topk_rs::Error::IoError`

topk-cli/src/util/files.rs
```diff
@@ -35,13 +36,22 @@ fn is_glob_pattern(pattern: &str) -> bool {
...
+}
+
 pub(crate) fn resolve_files(
     cwd: &Path,
     pattern: &str,
     recursive: bool,
 ) -> Result<Vec<UploadFile>, Error> {
     let path = expand_path(pattern, cwd)?;
 
+    path.try_exists().map_err(|e| file_access_error(&path, e))?;
```
petertoth: this does not need to be here as it's being checked in `collect_file()`

topk-cli/src/util/files.rs
```diff
@@ -35,13 +36,22 @@ fn is_glob_pattern(pattern: &str) -> bool {
     pattern.contains('*')
 }
 
+fn file_access_error(path: &Path, err: io::Error) -> Error {
+    Error::IoError(io::Error::new(
+        err.kind(),
+        format!("can't access file '{}': {}", path.display(), err),
+    ))
+}
```
petertoth: not needed as we can do conversion via `?`

jerguslejko:
ctx engine is gone :(

PR #446, opened 2026-05-03
[mcp] `include_content=True/False` and `get_content()`/`resources/read`

jerguslejko:


topk-cli/src/commands/ask.rs
```diff
@@ -146,6 +146,7 @@ pub async fn run(client: &Client, args: &AskArgs, output: &Output) -> Result<Ask
             None,
             args.mode.clone().map(Into::into),
             args.fields.clone(),
+            Some(true),
         )
         .await?;
```
petertoth, line 152: in `topk-cli`, the wrapper `SearchResult` should probably also have the `content_id` 

https://github.com/topk-io/topk/blob/cbc78155f54fddca0fcb672803bed227bdef675c/topk-cli/src/commands/search.rs#L19-L27

jerguslejko: yep, topk-py & topk-js also need updating, `Content` was previously always assumed to be there. can you take this change to topk-py/js/cli @petertoth ?

petertoth (approved):

PR #447, opened 2026-05-04
[topk-cli] fix output-dir references for json output

petertoth:


topk-cli/src/main.rs
```diff
@@ -237,25 +237,21 @@ async fn run(cli: Cli, output: &Output) -> Result<(), Error> {
             let client = make_client(&api_key, &region, &cli.host, cli.https);
 
             let result = ask::run(&client, &args, output).await?;
+            let paths = match args.output_dir.as_deref() {
+                Some(dir) => search::save_search_results(
+                    dir,
+                    result
```
jerguslejko: nit: just pass `&result.refs`

topk-cli/src/commands/search.rs
```diff
@@ -153,6 +153,26 @@ pub fn write_search_result(
     Ok(path.canonicalize().unwrap_or(path))
 }
 
+pub fn save_search_results<'a, I, S>(
```
jerguslejko: nit: use `for .. {}` style instead of iterators with side effects

topk-cli/src/main.rs
```diff
@@ -237,25 +237,21 @@ async fn run(cli: Cli, output: &Output) -> Result<(), Error> {
...
+                        .refs
+                        .iter()
+                        .map(|(ref_id, result)| (ref_id.as_str(), result)),
+                )?,
+                None => std::collections::HashMap::default(),
```
jerguslejko: import fully

topk-cli/src/commands/search.rs
jerguslejko: I'd consider merging this with `write_search_result` too

topk-cli/src/main.rs
```diff
@@ -284,25 +275,13 @@ async fn run(cli: Cli, output: &Output) -> Result<(), Error> {
             let client = make_client(&api_key, &region, &cli.host, cli.https);
 
             let result = search::run(&client, &args).await?;
+            let paths = match args.output_dir.as_deref() {
+                Some(dir) => search::save_search_results(dir, &result.refs())?,
```
jerguslejko: you're passing a `&result.refs()` anyway here, so no point in cloning it all in `pub fn refs(&self) -> HashMap<String, SearchResult> {`. 

Could just have `pub fn refs(&self) -> &HashMap<String, SearchResult>` and do the cloning

jerguslejko (approved):

PR #461, opened 2026-05-12
[docs] architecture guide

petertoth:


docs/architecture.mdx
```diff
@@ -6,18 +6,15 @@ icon: "layers"
 
 ## Overview
 
-TopK separates data read/write workloads from resource management across two operational planes: the **control plane** and the **data plane**.
-There is a single **global** control plane and multiple **regional** control planes, deployed alongside the regional data plane.
+TopK separates customer-data workloads from resource management across two operational planes: the **control plane** and the **data plane**.
```
jerguslejko, line 9: We are mixing the control/mgmt terminology in few places. I'd call it a "three layer system":

1. (Global) Management Plane - scope: orgs/projects; console.topk.io, api.topk.io for CLI (what we internally call "global control plane")
2. (Regional) Control Plane - scope: datasets/collections; api.{region}.topk.io
3. (Regional) Data Plane - scope: database/inference/ctx engine; api.{region}.topk.io, api.{region}.topk.io/mcp

docs/architecture.mdx
```diff
@@ -6,18 +6,15 @@ icon: "layers"
...
 
-This separation keeps account-level management out of the hot path for dataset reads, writes, retrieval, and inference, while allowing each regional deployment to operate close to the data it serves.
-
 <Info>
-  Dataset reads, writes, retrieval, and inference are served by the **regional data plane**.
-  The global control plane manages account-level resources and routing metadata; it does not serve customer-data workload requests.
+    This separation keeps account-level management out of the hot path for dataset reads, writes, retrieval, and inference, while allowing each regional deployment to operate close to the data it serves.
 </Info>
 
 ## Architecture principles
```
jerguslejko, line 20: I would replace this section with a diagram that shows the mgmt/control/data plane architecture described above. the persona here is a CTO/VP of eng/etc so they know what the principles are, we can show instead of tell

docs/architecture.mdx
```diff
@@ -6,18 +6,15 @@ icon: "layers"
...
+* The **global control plane** manages organizations, projects, API keys, account-level settings, and routing metadata. It also holds the region for each dataset and collection.
 * The **regional control plane** exposes management APIs for datasets and collections in that region.
 * The **regional data plane** runs customer-data workloads in the selected region, including storage, indexing, ingestion, retrieval, context workflows, and inference.
 
-This separation keeps account-level management out of the hot path for dataset reads, writes, retrieval, and inference, while allowing each regional deployment to operate close to the data it serves.
-
 <Info>
-  Dataset reads, writes, retrieval, and inference are served by the **regional data plane**.
-  The global control plane manages account-level resources and routing metadata; it does not serve customer-data workload requests.
+    This separation keeps account-level management out of the hot path for dataset reads, writes, retrieval, and inference, while allowing each regional deployment to operate close to the data it serves.
```
jerguslejko, line 17: Here I would drive the point that "db/storage, inference, and ctx engine are all in one region". and this can be either out public cloud, or a private deployment.

docs/architecture.mdx
jerguslejko, line 38: This is basically a repeat of the `# Overview`. I'd drop

jerguslejko, line 83: from customer point of view, console is part of the mgmt plane, not a client

jerguslejko, line 79: `Control API`. also can we show here that the regional control plane is a "proxy" for the global one - so reader doesn't get the impression that it's two separate databases

jerguslejko, line 73: `Management Plane`

jerguslejko, line 81: the boxes inside `Region` should be real things so `TopK Database`, `TopK Inference`, `TopK Context Engine`

jerguslejko: and once we do that, we don't need to have two arrows from (for example) SDKs to "regional data plane" and "regional management APIs", just one -> `Region`

jerguslejko, line 94: this section can be dropped IMO, it contains lil too much detail on the control plane and is mostly an enumeration of "Deployment model" ☝️

jerguslejko, line 180: we can present the database in more mature light. can reuse some older copy

jerguslejko, line 204: I'd merge these into a single box. From users point of view: processing cannot happen without an upload, persist cannot happen without processing

jerguslejko, line 209: same here

jerguslejko, line 234: Here maybe we should drop the diagram since it's not adding too much value. I'd just describe the interesting properties - it runs on in-cluster GPUs, right next to the database. Low latency, data residency.

jerguslejko, line 257: api keys are also in regional control plane used for auth, and _technically_ they are not in global, even though you can manage them in console.

jerguslejko, line 249: imo the most important part of "where the data lives" is not which resource lives in mgmt/control plane but where customer data lives. we should focus on that - db runs (in-region) on top of blob (in-region) and uses inference (in-region) which all power ctx engine (in-region)

jerguslejko, line 265: I'd drop this section. it follows from the system architecture above

jerguslejko, line 289: re-iteration of what was already said, I'd drop

jerguslejko, line 370: I'd drop

jerguslejko, line 307: we use "workflows" terminology on this page which we don't use anywhere else - I would remove the word completely

PR #467, opened 2026-05-19
[ctx] strong metadata checks

jerguslejko:


topk-rs/tests/test_dataset_delete.rs
```diff
@@ -71,28 +71,21 @@ async fn test_delete_document(ctx: &mut ProjectTestContext) {
...
-    // Deleting a non-existent document returns a handle
     let result = ctx
         .client
         .dataset(&dataset.name)
-        .wait_for_handle(&handle, None)
+        .delete("nonexistent")
         .await;
-    assert!(matches!(result, Ok(_)));
+
+    assert!(matches!(result, Err(Error::DatasetNotFound)));
```
marekgalovic, line 88: Should this be `DocumentNotFound`?

marekgalovic (commented):
We should use `DocumentNotFound` instead of `DatasetNotFound` for missing documents.

marekgalovic (approved):

topk-py/src/error.rs
```diff
@@ -57,6 +60,7 @@ create_exception!(error, CollectionNotFoundError, PyException);
 create_exception!(error, PartitionNotFoundError, PyException);
 create_exception!(error, DatasetAlreadyExistsError, PyException);
 create_exception!(error, DatasetNotFoundError, PyException);
+create_exception!(error, DocumentNotFoundError, PyException);
```
petertoth, line 63: for py we should add this exception to `.pyi` file and then run `make docs` https://docs.topk.io/sdk/topk-py/error

PR #470, opened 2026-05-20
[rs] structs test coverage

jerguslejko:
Add more e2e tests for check dotted notation in field names (eg. `meta.foo` vs `meta: { foo: ... }`).

### DIFF
--- a/topk-rs/tests/test_collection_struct.rs
+++ b/topk-rs/tests/test_collection_struct.rs
@@ -515,3 +515,67 @@ async fn test_struct_get_returns_flat_leaf(ctx: &mut ProjectTestContext) {
     assert!(one.get("meta").is_none());
     assert!(one.get("meta.year").is_none());
 }
+
+#[test_context(ProjectTestContext)]
+#[tokio::test]
+async fn test_struct_underscore_sub_field(ctx: &mut ProjectTestContext) {
+    let collection = ctx
+        .client
+        .collections()
+        .create(
+            ctx.wrap("test"),
+            HashMap::from_iter([(
+                "meta".to_string(),
+                FieldSpec::r#struct(false, [("_bar", FieldSpec::text(false, None))]),
+            )]),
+            None,
+        )
+        .await
+        .expect("could not create collection");
+
+    // Upsert document
+    let lsn = ctx
+        .client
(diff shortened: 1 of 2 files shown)


topk-rs/tests/test_collection_struct.rs
```diff
@@ -270,3 +270,312 @@ async fn test_struct_update(ctx: &mut ProjectTestContext) {
...
+        Some("alice"),
+    );
+    assert_eq!(meta.get("year").and_then(|v| v.as_i64()), Some(2024));
+    assert!(one.get("meta.author").is_none());
+    assert!(one.get("meta.year").is_none());
+}
+
+#[test_context(ProjectTestContext)]
+#[tokio::test]
+async fn test_struct_get_returns_whole_struct(ctx: &mut ProjectTestContext) {
```
marekgalovic, line 475: Isn't this the same as the test above?

topk-rs/tests/test_collection_struct.rs
```diff
@@ -270,3 +270,312 @@ async fn test_struct_update(ctx: &mut ProjectTestContext) {
...
+        .await
+        .expect("could not create collection");
+
+    // Upsert document
+    let lsn = ctx
+        .client
+        .collection(&collection.name)
+        .upsert(vec![doc!(
+            "_id" => "one",
+            "meta" => Value::r#struct([("_bar", "v".into())]),
```
marekgalovic, line 605: This might be a little confusing since we don't allow top-level fields to start with `_`. WDYT?

marekgalovic (approved):

topk-rs/tests/test_collection_struct.rs
jerguslejko: difference is how we select fields to _get_:

```
        .get(["one"], None, Some(lsn), None)
```

vs

```
        .get(["one"], Some(vec!["meta".to_string()]), Some(lsn), None)
```

topk-rs/tests/test_collection_struct.rs
jerguslejko: Agno specifically does this as `meta_data._agno`. Technically no reason for us to prevent this.

PR #483, opened 2026-06-01
partitions

marekgalovic:
Expose partitioned collections through Rust SDK.


topk-rs/tests/test_partitions.rs
```diff
@@ -0,0 +1,935 @@
...
+
+    let err = ctx
+        .client
+        .collection(&collection.name)
+        .partition("missing-partition")
+        .count(None, None)
+        .await
+        .expect_err("should not be able to query a partition that was never created");
+
+    assert!(matches!(err, Error::CollectionNotFound));
```
jerguslejko: @marekgalovic should we return `PartitionNotFound` here?

jerguslejko (approved):

PR #494, opened 2026-06-12
[sql] integrate CI 

jerguslejko:


topk-sql/src/ext/stmt.rs
```diff
@@ -37,17 +37,41 @@ impl SqlStatementExt for SqlStatement {
         count
     }
 
-    fn table_name(&self) -> Option<String> {
+    fn table_ref(&self) -> Option<(Option<&str>, &str)> {
```
jerguslejko, line 46: unrelated change to allow passing back `<schema>.<collection>`

PR #495, opened 2026-06-15
[sql] custom syntax for `INDEX`es

jerguslejko:
Replace `CREATE TABLE + CREATE INDEX` statement bundling with a native syntax for defining an index on a column:

```sql
CREATE TABLE books (
    title          TEXT NOT NULL               INDEX keyword_index(),
    bio            TEXT                        INDEX semantic_index(),
    embedding      f32_vector(4)               INDEX vector_index(metric = 'cosine'),
    sparse_emb     f32_sparse_vector           INDEX vector_index(metric = 'dot_product'),
    multi_emb      f32_matrix(4)               INDEX multi_vector_index(metric = 'maxsim'),
);
```

### DIFF
--- a/topk-sql/src/stmt/create_table.rs
+++ b/topk-sql/src/stmt/create_table.rs
@@ -1,8 +1,8 @@
 use std::collections::HashMap;
 
 use sqlparser::ast::{
-    ArrayElemTypeDef, BinaryOperator, ColumnDef, ColumnOption, CreateIndex as SqlCreateIndex,
-    CreateTable as SqlCreateTable, DataType, Expr as SqlExpr,
+    ArrayElemTypeDef, BinaryOperator, ColumnDef, ColumnOption, CreateTable as SqlCreateTable,
+    DataType, Expr as SqlExpr, Function, FunctionArg, FunctionArgExpr, FunctionArguments,
 };
 use topk_rs::proto::v1::control::{
     FieldIndex, FieldSpec, FieldType, KeywordIndexType, MultiVectorDistanceMetric,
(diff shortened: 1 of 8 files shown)


topk-sql/src/stmt/create_table.rs
```diff
@@ -44,124 +44,20 @@ impl TryFrom<SqlCreateTable> for Statement {
...
-        sql_invalid!(
-            idx.columns.len() != 1,
-            "CREATE INDEX must reference exactly one column"
-        );
-
-        // Parse table name.
-        let table = Table::new(idx.table_name)?;
-        sql_invalid!(
-            !matches!(table, Table::Collection(_)),
-            "CREATE INDEX requires a collection name"
```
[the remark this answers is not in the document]
jerguslejko: That's on purpose, we're defining a custom dialect.

topk-sql/src/stmt/create_table.rs
```diff
@@ -44,124 +44,20 @@ impl TryFrom<SqlCreateTable> for Statement {
...
         let data_type = FieldType::from_sql(column.data_type)?;
 
         let mut required = false;
+        let mut index = None;
         for option in column.options {
             match option.option {
                 ColumnOption::Null => {}
                 ColumnOption::NotNull => required = true,
+                ColumnOption::Check(SqlExpr::Function(func)) => {
+                    sql_invalid!(index.is_some(), "column has multiple INDEX definitions");
```
marekgalovic, line 58: This is not possible with the new syntax, right?

marekgalovic (approved):

topk-sql/src/stmt/create_table.rs
jerguslejko: we override `ColumnOption` parser which does allow multiple column options, so `title TEXT NOT NULL INDEX keyword_index() INDEX keyword_index()` would parse. This is a check against that.

PR #497, opened 2026-06-15
[docs] sql

petertoth:


topk-sql/README.md
```diff
@@ -109,18 +153,12 @@ CREATE TABLE [IF NOT EXISTS] <table> (
...
-| `semantic_index` | `TEXT`, `VARCHAR` | — |
-| `vector_index` | all `*_vector(n)` and `*_sparse_vector` types | `metric` |
-| `multi_vector_index` | all `*_matrix(n)` types | `metric`, `quantization`, `width`, `top_k` |
-
-`metric` values for `vector_index`: `cosine`, `dot_product`, `euclidean`, `hamming`.
-
-`metric` values for `multi_vector_index`: `maxsim`.
-
-`quantization` values: `1bit`, `2bit`, `scalar`.
+| Method | Applies to | Options |
```
jerguslejko: the options render a bit weird with the `|`

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+
+TopK collections are schemaless by default. Documents can have any fields with any
+types — no schema declaration is required to store or query them.
+
+The only exception: **fields you want to index must be declared**, because index
+configuration (metric, index type, etc.) is attached to the field definition.
+
+<Tip>
+
+Most columns arrive as JSON. Drivers deserialize them automatically (`psycopg2` → `dict`/`list`, `node-postgres` → object). To get a typed column, use the `::` cast operator.
```
jerguslejko: not true - we try to infer types as much as possible - I wouldn't lead with this. Basically we try to infer the type from schema, statically (eg. return values of search fns), from casts, and only then fall back to JSON - for cases such as having `1` and `"foo"` in the same column. Lets reword this as JSON is the escape hatch but by default we send the native type.

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+
+```sql
+SELECT title, published_year::int4, rating::float8 FROM books LIMIT 10;
+```
+
+## SQL Reference
+
+### Table references
+
+DML statements use `<table>` as either a collection name or a collection plus partition:
```
jerguslejko: drop `DML`? I assume most people won't be familiar with the shortcut

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+| `schema.collection` | `public.books` |
+| `collection$partition` | `books$2024` |
+| `schema.collection$partition` | `public.books$2024` |
+| `collection PARTITION name` | `books PARTITION 2024` |
+
+Partition syntax applies to `SELECT`, `INSERT`, `UPDATE`, and `DELETE` only. DDL
+(`CREATE TABLE`, `DROP TABLE`) names collections without partitions;
+partitions are created implicitly on first write.
+
+### CREATE TABLE
```
jerguslejko: lets start with `SELECT`, then `INSERT/`UPDATE`/`DELETE` and then `CREATE/DROP TABLE`?

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+```sql
+DELETE FROM <table> WHERE _id = '<id>';
+DELETE FROM <table> WHERE _id IN ('<id1>', '<id2>', ...);
+DELETE FROM <table> WHERE <filter_expr>;
+DELETE FROM <collection>$<partition>;
+```
+
+### SELECT
+
+```sql
```
jerguslejko: would be nice to start with some prose

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+SELECT <projection> FROM <table>
+  [WHERE <filter>]
+  [ORDER BY <expr> [ASC | DESC]]
+  [LIMIT <n>]
```
jerguslejko: technically this is `[ORDER BY <expr> [ASC | DESC] LIMIT <n>`

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+```
+
+`OFFSET` is not supported — use `LIMIT` only.
```
jerguslejko: wdyt about moving this to a single "Not supported" section where we can give some motivation for why they are not supported. `OFFSET` just doesn't make sense in a search db. multiple `ORDER BY`s are compensated with a scoring expr. `SELECT *` is just anti-pattern, it pulls all the data, etc..

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+Each item in the select list must be one of:
+
+- A named column: `SELECT title, rating`
+- An aliased expression: `SELECT rating * 2 AS double_rating`
+- A search function: `SELECT vector_distance(embedding, '[1,0,0,0]'::f32_vector) AS score`
+- A wire-type cast: `SELECT rating::float8 AS rating_f64`
+
+<Warning>
+
+Limitations on the SELECT list:
```
jerguslejko: same as above - wdyt about moving these to Postgres Compatibility section or something like that?

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+- `SELECT *` is not supported — list columns explicitly.
+- Indexed vector fields (`embedding`, `sparse_emb`, …) cannot be selected directly — use search functions (`vector_distance`, …) instead.
+- `COUNT(*)` cannot be combined with other columns in the same `SELECT` list.
+
+</Warning>
+
+Struct subfields (`metadata.publisher`) can be selected; whole struct columns cannot.
+
+Wire-type casts (`::int4`, `::float8`, `::text`, …) are accepted in the SELECT list
+and affect `RowDescription` OIDs only — see [Output type mapping](#output-type-mapping).
```
jerguslejko: we just kinda start talking about `RowDescription` here with no context

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+|----------|---------|
+| `=`, `<>`, `!=`, `<`, `<=`, `>`, `>=` | `rating > 4.0` |
+| `AND`, `OR`, `NOT` | `genre = 'fantasy' AND in_print = true` |
+| `IS NULL`, `IS NOT NULL` | `checksum IS NOT NULL` |
+| `IN`, `NOT IN` | `genre IN ('fantasy', 'fiction')` |
+| `BETWEEN`, `NOT BETWEEN` | `published_year BETWEEN 1900 AND 2000` |
+| `LIKE`, `NOT LIKE` | `title LIKE 'The%'` |
+| `~` (regex) | `author ~ 'Tol.*'` |
+| Arithmetic | `rating * 10 > 40` |
+| `CASE WHEN … THEN … ELSE … END` | — |
```
jerguslejko: missing example

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+## Type System
+
+TopK-native literals use **`::topk_type` casts** (preferred in `INSERT`/`UPDATE`
+`VALUES` and search-function arguments) or equivalent **constructor calls**.
+PostgreSQL wire-type casts (`::float8`, …) apply only in `SELECT` projection.
+
+| Constructor | Example |
+|-------------|---------|
+| `f32_vector(ARRAY[…])` | `'[0.1, 0.2, 0.3]'::f32_vector` or `f32_vector(ARRAY[0.1, 0.2, 0.3])` |
+| `f16_vector(ARRAY[…])` | (same pattern as f32) |
```
jerguslejko: nit: drop `... as f32`?

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+| `f8_vector(ARRAY[…])` | (same pattern) |
+| `u8_vector(ARRAY[…])` | (same pattern) |
+| `i8_vector(ARRAY[…])` | (same pattern) |
+| `binary_vector(ARRAY[…])` | (same pattern) |
+| `f32_sparse_vector(ARRAY[idx], ARRAY[val])` | `f32_sparse_vector(ARRAY[0, 2], ARRAY[1.0, 0.5])` |
+| `f32_sparse_vector(JSON)` | `'{"0": 1.0, "2": 0.5}'::f32_sparse_vector` |
```
jerguslejko: nit: example in one line?

jerguslejko: btw we also support https://github.com/topk-io/topk/blob/3f0c0995cf24b777cfce934cbb3c0d365fbbfa2b/topk-sql/README.md#L414 syntax

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+| `f8_matrix(…)` | (same pattern) |
+| `u8_matrix(…)` | (same pattern) |
+| `i8_matrix(…)` | (same pattern) |
+| `bytes('hexstring')` | `bytes('deadbeef')` |
+| `struct(key1, val1, …)` | `struct('publisher', 'Penguin', 'pages', 320)` |
+| `ARRAY[elem, …]` | `ARRAY['classic', 'fiction']` (list) |
+
+### Complex types → JSON
+
+No PostgreSQL OID exists for TopK-native vector/matrix/sparse/list types. They are
```
jerguslejko: that's true but we also don't recommend users pull these fields out - for indexed fields, that's not possible. WDYT we rephrase this section so it seems more like a design decision and not a limitation

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+| `::real` / `::float4` | 700 `FLOAT4` |
+| `::float` / `::float8` | 701 `FLOAT8` |
+| `::text` | 25 `TEXT` |
+| `::bytea` | 17 `BYTEA` |
+| `::json` / `::jsonb` | 114 `JSON` |
+| plain column (no cast) | 114 `JSON` |
+| search function (no cast) | 700 `FLOAT4` |
+| `COUNT(*)` (no cast) | 20 `INT8` |
+
+## Search Functions
```
jerguslejko: this is meat of SQL tbh - we should put this under `SELECT` (which should be first) so it's visible and easily findable

docs/sdk/topk-sql/overview.mdx
```diff
@@ -0,0 +1,565 @@
...
+
+Scoring functions go in `SELECT` only — alias them, then reference the alias in `ORDER BY`. Filter functions go in `WHERE`.
+
+| Function | Use in | Description |
```
jerguslejko: also, maybe we should split this table into subtables - distance fns & bm25_score are scoring fns, match_all/match_any are predicates

jerguslejko (approved):

PR #498, opened 2026-06-17
[topk-cli] infer MIME type from symlink target on upload

petertoth:


jerguslejko (changes requested):
we shouldn't reject symlinks. it's very common to have them and it's a UX pain to work around this limitation from the user point of view

petertoth:
> we shouldn't reject symlinks. it's very common to have them and it's a UX pain to work around this limitation from the user point of view

updated to just infer the mime type from the target file if symlink

jerguslejko:
ctx engine is gone :(

PR #500, opened 2026-06-25
[sql] `pg_catalog`

jerguslejko:
* upgrade `sqlparser` from `0.52` to `0.61`
 * cleanup tests around `pg_catalog`

`test_info` (pg_catalog) tests will pass after upstream is deployed.


topk-sql/src/expr/value.rs
```diff
@@ -23,7 +23,13 @@ impl FromSql<SqlExpr> for Value {
                 op: UnaryOperator::Minus,
                 expr,
             } => match *expr {
-                SqlExpr::Value(SqlValue::Number(repr, _)) => parse_number(&format!("-{repr}")),
+                SqlExpr::Value(v) if matches!(v.value, SqlValue::Number(_, _)) => {
+                    let repr = match v.value {
+                        SqlValue::Number(repr, _) => repr,
+                        _ => unreachable!(),
```
petertoth, line 29: should be error

topk-sql/src/ext/stmt.rs
```diff
@@ -35,42 +39,58 @@ impl SqlStatementExt for SqlStatement {
...
+                    Some(name) => name,
+                    None => return Ok(None),
+                }
             }
-            _ => None,
-        }
+            _ => return Ok(None),
+        };
+
+        let table = Table::new(name.clone())?;
```
petertoth, line 81: nit: inline

petertoth (approved):

topk-sql/src/expr/value.rs
jerguslejko: we already check that i's a number above

PR #504, opened 2026-07-03
[rs] `topk_rs::Value` <-> `serde_json::Value`

jerguslejko:
The JSON mapping is useful on it's own, pulling out the trait.


topk-rs/src/proto/data/v1/data_ext/value.rs
```diff
@@ -349,6 +349,267 @@ impl Value {
...
+
+    fn try_from(value: serde_json::Value) -> Result<Self, Self::Error> {
+        match value {
+            serde_json::Value::Null => Ok(Value::null()),
+            serde_json::Value::Bool(value) => Ok(Value::bool(value)),
+            serde_json::Value::String(value) => Ok(Value::string(value)),
+            serde_json::Value::Number(value) => {
+                if let Some(value) = value.as_i64() {
+                    Ok(Value::i64(value))
+                } else if let Some(value) = value.as_u64() {
```
marekgalovic: IMO we should try u32/f32 first to save storage, and if that doesn't fit, bump to u64/f64.

topk-rs/src/proto/data/v1/data_ext/value.rs
```diff
@@ -349,6 +349,267 @@ impl Value {
...
+            serde_json::Value::Object(value) => value
+                .into_iter()
+                .map(|(key, value)| Ok((key, Value::try_from(value)?)))
+                .collect::<Result<HashMap<_, _>, _>>()
+                .map(Value::r#struct),
+            serde_json::Value::Array(value) => {
+                if value.is_empty() {
+                    return Ok(Value::list(Vec::<f32>::new()));
+                }
+                if value.iter().all(serde_json::Value::is_number) {
```
marekgalovic: What about all integer arrays? Ie. if all values are `[0, 255]` we can map to `Vec<u8>`. We should also use the "smallest numeric type" possible similar to scalars.

marekgalovic (approved):

PR #510, opened 2026-07-10
[rs] `should` text expression

jerguslejko:
`should(): TextExpr` allows _scoring without filtering_.

Previously, it was impossible to get a text score for a document without excluding documents that don't match the text query from the result set. 

### Example

```rust
// Gate the result set on "love", but let "wealth" boost the score of
// documents that also mention it — without excluding those that don't:
collection.query(
    select([("bm25", fns::bm25_score(None, None))])
        .filter(
            r#match("love", Some("summary"), None, false)
                .and(should("wealth", Some("summary"), None)),
        )
        .sort(field("bm25"), false)
        .limit(10),
    None,
    None,
)
```

A `should`-only filter matches the entire collection — it ranks documents by BM25 without excluding anything.

### DIFF
--- a/topk-rs/tests/test_query_text.rs
+++ b/topk-rs/tests/test_query_text.rs
@@ -4,7 +4,7 @@ use topk_rs::{
     data::literal,
     doc,
     proto::v1::control::{FieldIndex, FieldSpec, KeywordIndexType},
-    query::{field, filter, fns, r#match, select},
+    query::{field, filter, fns, r#match, select, should},
     schema, Error,
 };
 use utils::dataset;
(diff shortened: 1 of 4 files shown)


topk-rs/tests/test_query_text.rs
```diff
@@ -125,6 +125,56 @@ async fn test_query_text_filter_stop_word(ctx: &mut ProjectTestContext) {
...
 
+#[test_context(ProjectTestContext)]
+#[tokio::test]
+async fn test_query_text_should_does_not_filter(ctx: &mut ProjectTestContext) {
+    let collection = dataset::books::setup(ctx).await;
+
+    let result = ctx
+        .client
+        .collection(&collection.name)
+        .query(
```
marekgalovic, line 136: add select for BM25 score here to assert that it scores without filtering

marekgalovic (approved):

PR #513, opened 2026-07-10
[topk-py][topk-js][topk-sql] offset

petertoth:


topk-py/tests/test_query_limit.py
```diff
@@ -79,6 +79,46 @@ def test_query_limit_vector_distance(ctx: ProjectContext):
     assert docs_limit == docs_topk
 
 
+def _setup_books_chunked(ctx: ProjectContext):
+    # Upsert in chunks (rather than dataset.books.setup's single bulk upsert)
+    # so the collection has multiple segments, matching the topk-rs test
+    # fixture. The limit+offset collector merge needs that to compute the
+    # right window; a single-segment collection returns the wrong window.
```
jerguslejko: what? that doesn't seem right or we have a bug somewhere

topk-sql/src/stmt/select.rs
```diff
@@ -141,21 +141,27 @@ impl TryFrom<SqlQuery> for Statement {
...
-                    limit: Some(ref expr),
-                    ..
-                }) => Some(expr.as_u64().ok_or_else(|| {
-                    Error::Invalid("LIMIT must be a positive integer".to_string())
-                })?),
-                Some(_) => sql_invalid!("LIMIT must be a positive integer"),
-                None => None,
-            };
+        let (limit, offset) = match query.limit_clause {
+            Some(LimitClause::OffsetCommaLimit { .. }) => sql_unsupported!("LIMIT offset, limit"),
```
jerguslejko: what are the semantics of `LIMIT offset, limit`?

topk-sql/src/stmt/select.rs
```diff
@@ -164,9 +170,16 @@ impl TryFrom<SqlQuery> for Statement {
             }
             (Some(_), None) => sql_invalid!("ORDER BY without LIMIT is not supported"),
             (None, Some(k)) => stages.push(Stage::limit(k)),
+            (None, None) if offset.is_some() => {
+                sql_invalid!("OFFSET without LIMIT is not supported")
+            }
             (None, None) => {}
```
jerguslejko: personal nit: I'd collapse these two cases and push the `if` inside

topk-sql/src/stmt/select.rs
petertoth: `LIMIT offset, limit` is mysql shorthand for `LIMIT 5 OFFSET 10`
but it's not postgres compatible

jerguslejko: I see. Does this syntax actually pass through the parser? The `sqlparser` ast is a union of all supported dialects.  We don't assert "unsupported" for non-postgres features since it fails during parsing (doesn't get to this point)

topk-sql/tests/test_select.rs
```diff
@@ -790,7 +812,14 @@ async fn semantic_similarity_search() {
...
 )]
-#[case::offset("SELECT _id FROM {{table}} LIMIT 5 OFFSET 3", "Unsupported: OFFSET")]
+#[case::offset_without_limit(
+    "SELECT _id FROM {{table}} OFFSET 3",
+    "Invalid: OFFSET without LIMIT is not supported"
+)]
+#[case::offset_comma_limit(
+    "SELECT _id FROM {{table}} LIMIT 4, 3",
+    "Parse error: sql parser error: Expected: end of statement, found: , at Line: 1, Column: 64"
+)]
```
jerguslejko: we can also drop this test IMO, since it's mysql syntax

jerguslejko (approved):

PR #514, opened 2026-07-10
[topk-rs] Multi-expr sort

marekgalovic:
Add support for sorting with multiple exprs (fields). Each expr can have separate asc/desc order.

Supported `sort` call signatures:
```rust
// Sort by field "foo", DESC
.sort("foo")
// Sort by expr, DESC
.sort(expr) 
// Sort by expr, with specific order (ASC)
.sort((expr, SortOrder::Asc)) 
// Sort by exprs, each with their respecitve order
.sort([(expr, SortOrder::Asc), (expr2, SortOrder::Desc), ...])
```


topk-rs/src/proto/data/v1/query_ext/mod.rs
```diff
@@ -29,7 +33,10 @@ impl Query {
     /// Use `.sort(expr, asc).limit(k)` instead.
     #[deprecated(note = "Use `.sort(expr, asc).limit(k)` instead")]
     pub fn topk(mut self, expr: LogicalExpr, k: u64, asc: bool) -> Self {
-        self.stages.push(Stage::sort(expr, asc));
+        self.stages.push(Stage::sort([(
+            expr,
+            asc.then_some(SortOrder::Asc).unwrap_or(SortOrder::Desc),
```
jerguslejko, line 38: `if asc { SortOrder::Asc } else { SortOrder::Desc }`?

jerguslejko (approved):

PR #518, opened 2026-07-15
[topk-rs] json `Value`

jerguslejko:
Complementary to https://github.com/topk-io/topk/pull/504

The proto `Value` already has a derived serde impl that emits protobuf-shaped JSON (`{"value": {"String": "a"}}`) and can't be overridden — the `json::Value` newtype carries the natural-JSON `Serialize`/`Deserialize` so structs (like `SearchResult`) can just derive serde.

### DIFF
--- a/topk-rs/src/json.rs
+++ b/topk-rs/src/json.rs
@@ -155,33 +155,17 @@ impl TryFrom<Vec<Vec<serde_json::Value>>> for TopkValue {
             }
         }
 
-        if rows.iter().map(Vec::len).sum::<usize>() >= u32::MAX as usize {
+        if rows.len() * num_cols >= u32::MAX as usize {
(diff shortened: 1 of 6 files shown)


topk-rs/src/json.rs
```diff
@@ -281,6 +321,56 @@ impl TryFrom<Value> for serde_json::Value {
...
+    if !(value.is_finite() && value.fract() == 0.0) {
+        return Ok(None);
+    }
+
+    if value >= i64::MIN as f64 && value <= i64::MAX as f64 {
+        Ok(Some(value as i64))
+    } else {
+        Ok(None)
+    }
+}
```
petertoth: this could be simplified / better described

petertoth (approved):

PR #524, opened 2026-07-23
Release 0.13.0 & group by + offset docs  

petertoth:


jerguslejko (commented):
@petertoth we are missing `.group_by()`/`.offset()` docs in https://docs.topk.io/collections/query too

petertoth:
> @petertoth we are missing `.group_by()`/`.offset()` docs in https://docs.topk.io/collections/query too

this PR is adding them to `https://docs.topk.io/collections/query`

i will also add offset

jerguslejko (approved):

PR #534, opened 2026-07-29
[py/js] unknown proto variants

jerguslejko:
Old SDKs abort on data from a newer server — prost drops unknown fields, the oneof decodes as `None`, the exhaustive match panics. topk-py was worse: it mapped that `None` to `Value::Null()`, so read-modify-write silently nulled the field (upsert is full-replace).

Adds `Unknown` to `Value`, `DataType` and `FieldIndex`; sub-enums collapse into them. Values project to natives, so they get an `UnknownValue` sentinel, rejected on write. Decode no longer panics; write path unchanged.

### DIFF
--- a/topk-js/src/schema/field_index.rs
+++ b/topk-js/src/schema/field_index.rs
@@ -1,6 +1,14 @@
 use napi::bindgen_prelude::*;
 use napi_derive::napi;
 
+use crate::data::UNSUPPORTED;
+use topk_rs::proto::v1::control::{
+    KeywordIndexType as KeywordIndexTypePb,
+    MultiVectorDistanceMetric as MultiVectorDistanceMetricPb,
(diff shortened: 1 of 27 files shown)


topk-js/src/schema/field_index.rs
```diff
@@ -1,6 +1,14 @@
 use napi::bindgen_prelude::*;
 use napi_derive::napi;
 
+use crate::data::UNSUPPORTED;
+use topk_rs::proto::v1::control::{
+    KeywordIndexType as KeywordIndexTypePb,
+    MultiVectorDistanceMetric as MultiVectorDistanceMetricPb,
+    MultiVectorQuantization as MultiVectorQuantizationPb,
+    VectorDistanceMetric as VectorDistanceMetricPb,
+};
```
petertoth, line 10: nit: imports

petertoth (approved):

PR #542, opened 2026-08-05
aloe asset

jlsan92:


jerguslejko (commented):
@jlsan92 wanna compress that image? i think pilea img is also huge

petertoth:
> @jlsan92 wanna compress that image? i think pilea img is also huge

pretty good tool for it https://compressor.io/

jerguslejko (approved):

PR #544, opened 2026-08-05
Release 0.14.0

petertoth:


topk-js/index.js
```diff
@@ -77,8 +77,8 @@ function requireNative() {
       try {
         const binding = require('topk-js-android-arm64')
         const bindingPackageVersion = require('topk-js-android-arm64/package.json').version
-        if (bindingPackageVersion !== '0.13.0' && process.env.NAPI_RS_ENFORCE_VERSION_CHECK && process.env.NAPI_RS_ENFORCE_VERSION_CHECK !== '0') {
```
jerguslejko, line 80: @petertoth can we pull out the version into a var/const on top?

jerguslejko (approved):

topk-js/index.js
petertoth: this is a generated file

PR #551, opened 2026-08-09
[tests] delete filter must be boolean

jerguslejko:
adding missing tests

### DIFF
--- a/topk-rs/tests/test_collection_delete.rs
+++ b/topk-rs/tests/test_collection_delete.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 use test_context::test_context;
 use topk_rs::data::literal;
 use topk_rs::doc;
+use topk_rs::proto::v1::control::FieldSpec;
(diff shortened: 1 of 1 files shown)


topk-rs/tests/test_collection_delete.rs
```diff
@@ -289,16 +290,90 @@ async fn test_delete_with_invalid_filter(ctx: &mut ProjectTestContext) {
         assert_eq!(lsn, format!("{}", batch_idx + 1));
     }
 
+    // None of these are boolean predicates: a non-boolean operand, a known
+    // bare field, and an unknown bare field.
+    for expr in [
+        field("batch_idx").gte(literal(1)).or(field("batch_idx")),
```
marekgalovic: This should fail to compile so won't even hit the result type validation

marekgalovic (approved):

topk-rs/tests/test_collection_delete.rs
jerguslejko: Fair, i'll split.

PR #567, opened 2026-08-24
[sdk] derive e2e endpoints, move cli tests

jerguslejko:
cli e2e tests move from `src/commands` to `topk-cli/tests/`; e2e endpoints derived from region + host instead of hardcoded per-service hosts.


topk-js/.yarnrc.yml
petertoth, line 1: what is this doing ?

petertoth (approved):

PR #582, opened 2026-08-31
[sdk] functions in logical exprs

jerguslejko:
Make it possible to use functions in logical expr positions. This makes expressing `SELECT _id FROM table ORDER BY vector_distance(...)`-style queries natural in topk-sql/topk-es, but also in topk-rs/js/py.

`fns::vector_distance()` still returns `FunctionExpr`, and we use `lift!` macros to mirror logical combinators so the two forms can be used interchangeably.  

```
fns::vector_distance() -> FunctionExpr
fns::vector_distance().add(1) -> LogicalExpr
```


topk-js/src/expr/logical/mod.rs
```diff
@@ -35,16 +36,22 @@ use stringy::Stringy;
...
         value: napi::sys::napi_value,
     ) -> napi::Result<Self> {
-        let expr = crate::try_cast_ref!(env, value, LogicalExpression)?;
-        Ok(expr.clone())
+        if let Ok(expr) = crate::try_cast_ref!(env, value, LogicalExpression) {
+            return Ok(expr.clone());
+        }
+
+        // Accept a score function wherever a logical expression is expected,
+        // e.g. `.topk(fn.bm25Score(), 10)`.
```
petertoth: `.sort(...)` ?

topk-rs/src/proto/data/v1/query_ext/expr_ext/function.rs
```diff
@@ -90,3 +90,44 @@ impl FunctionExpr {
...
+    ($($fn:ident($($arg:ident: $ty:ty),*)),* $(,)?) => {
+        impl FunctionExpr {
+            $(pub fn $fn(self, $($arg: $ty),*) -> LogicalExpr {
+                LogicalExpr::function(self).$fn($($arg),*)
+            })*
+        }
+    };
+}
+
+lift!(
```
marekgalovic, line 104: nit - would be nice if we didn't have to keep this in sync with every expr we add into `LogicalExpr`.

marekgalovic (approved):

topk-rs/src/proto/data/v1/query_ext/expr_ext/function.rs
jerguslejko: in Rust we could do something like:

```rs
pub trait ExprOps: Into<LogicalExpr> + Sized {
    fn eq(self, right: impl Into<LogicalExpr>) -> LogicalExpr {
        LogicalExpr::binary(binary_op::Op::Eq, self, right)
    }
}

impl ExprOps for LogicalExpr {}
impl ExprOps for FunctionExpr {}
```

but that doesn't work in py/js. `lift!` is the most "uniform" way to implement lifting across languages afaik.

Another alternative is change the return type of `fn.bm25_score()` to produce `LogicalExpr` what that breaks in Rust since we have [methods on fn exprs](https://github.com/topk-io/topk/blob/2ad4c5a6626aa858f019e92960a7996b6a44d333/topk-rs/src/proto/data/v1/query_ext/expr_ext/function.rs#L81C16-L81C16)

marekgalovic (approved):
proto/rs changes LGTM. @petertoth to 👍 py/js/docs/...

topk-js/src/expr/sort.rs
```diff
@@ -31,24 +31,32 @@ impl From<SortOrder> for topk_rs::proto::v1::data::stage::sort_stage::SortOrder
...
 
-#[derive(Debug, Clone)]
-pub struct SortExpression {
-    pub expr: LogicalExpression,
-    pub order: SortOrder,
+/// Either a single expression to sort by, or an array of `(expr, order)` pairs.
+pub enum SortArg {
+    Single(LogicalExpression),
+    Many(Vec<SortExpr>),
 }
```
petertoth, line 47: Should we keep the `SortExpression` so it's in line with JS naming (`LogicalExpression`, `FilterExpression`) ?
PY is using `...Expr`

topk-js/src/expr/filter.rs
```diff
@@ -17,31 +12,23 @@ impl FromNapiValue for FilterExpression {
...
-            });
+            return Ok(FilterExpression::Text { expr: expr.clone() });
         }
 
-        Err(napi::Error::from_reason(
-            "Unsupported filter expression value",
-        ))
+        let expr = unsafe { LogicalExpression::from_napi_value(env, value) }
+            .map_err(|_| napi::Error::from_reason("Unsupported filter expression value"))?;
+        Ok(FilterExpression::Logical { expr })
```
petertoth: Why was `crate::try_cast_ref!(env, value, LogicalExpression)` replaced with `unsafe { LogicalExpression::from_napi_value(env, value) }` ?

petertoth (approved):

topk-js/src/expr/filter.rs
jerguslejko: reverted

topk-js/src/expr/sort.rs
jerguslejko: discussed in person, it was `SortExpr` already

PR #565, opened 2026-08-24
[es] date math, histogram, time zones

jerguslejko:

PR #583, opened 2026-09-01
[docs] document the Elasticsearch API

davidmyriel:
TopK serves an Elasticsearch-compatible endpoint at <region>.es.topk.io. Any official Elasticsearch client talks to it unchanged, but nothing documented it. Since topk-es is publish = false, the only sign it existed was 4.9k lines of Rust.

Every claim was checked against a live cluster, which corrected several things reading the source alone had got wrong.

### DIFF
--- a/docs/sdk/topk-es/overview.mdx
+++ b/docs/sdk/topk-es/overview.mdx
@@ -63,7 +63,7 @@ PUT /books
       "published_year": { "type": "integer" },
       "rating":         { "type": "float" },
       "in_print":       { "type": "boolean" },
-      "embedding":      { "type": "dense_vector", "dims": 16, "similarity": "cosine" }
+      "embedding":      { "type": "dense_vector", "dims": 4, "similarity": "cosine" }
     }
   }
 }
@@ -74,9 +74,9 @@ Index documents:
 ```json
 POST /books/_bulk
 { "index": { "_id": "mockingbird" } }
-{ "title": "To Kill a Mockingbird", "author": "Lee", "published_year": 1960, "rating": 4.3, "in_print": true }
+{ "title": "To Kill a Mockingbird", "author": "Lee", "published_year": 1960, "rating": 4.3, "in_print": true, "embedding": [1.0, 0.0, 0.0, 0.0] }
 { "index": { "_id": "gatsby" } }
-{ "title": "The Great Gatsby", "author": "Fitzgerald", "published_year": 1925, "rating": 3.9, "in_print": true }
+{ "title": "The Great Gatsby", "author": "Fitzgerald", "published_year": 1925, "rating": 3.9, "in_print": true, "embedding": [0.9, 0.1, 0.0, 0.0] }
 ```
 
 Search:
@@ -152,7 +152,7 @@ The following query clauses are supported inside `query`:
 | `regexp` | Regular expression match. |
 | `range` | `gt`, `gte`, `lt`, `lte` bounds. |
 | `exists` | Field is present and non-null. |
-| `bool` | Combine clauses with `must`, `should`, `must_not`, `filter`, `minimum_should_match`. |
+| `bool` | Combine clauses with `must`, `should`, `must_not`, `filter`. Accepts `boost`. |
 | `semantic` | **TopK extension.** Managed semantic search — see below. |
 
 ## Search body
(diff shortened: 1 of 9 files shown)


docs/sdk/topk-es/overview.mdx
```diff
@@ -0,0 +1,364 @@
...
+
+`knn` takes `field`, `query_vector`, and `k`, plus optional `filter`, `num_candidates`,
+`boost`, and `similarity`.
+
+```json
+POST /books/_search
+{
+  "knn": {
+    "field": "embedding",
+    "query_vector": [0.12, 0.87, 0.05, 0.44],
```
jerguslejko: all query vector examples use 4 dims, but the schema above defines `dims: 16`. Should align to be copy-pastable

topk-es/README.md
```diff
@@ -0,0 +1,360 @@
...
+| `match` | Full-text match. Accepts `query`, `operator` (`or` / `and`), `boost`. |
+| `multi_match` | Match across several fields. |
+| `term` | Exact term match. |
+| `terms` | Match any of a list of terms. |
+| `ids` | Match documents by `_id`. |
+| `prefix` | Prefix match on a string field. |
+| `regexp` | Regular expression match. |
+| `range` | `gt`, `gte`, `lt`, `lte` bounds. |
+| `exists` | Field is present and non-null. |
+| `bool` | Combine clauses with `must`, `should`, `must_not`, `filter`, `minimum_should_match`. |
```
jerguslejko: `minimum_should_match` actually wasn't implemented, adding in https://github.com/topk-io/topk/pull/585

topk-es/README.md
```diff
@@ -0,0 +1,360 @@
...
+| `400` | `action_request_validation_exception` | The request was structurally invalid — for example an unknown mapping type. |
+| `400` | `invalid_index_name_exception` | Index name violates the naming rules. |
+| `400` | `invalid_document_id_exception` | Document id is empty or over 512 bytes. |
+| `400` | `resource_already_exists_exception` | The index already exists. |
+| `400` | `json_parse_exception` | The body was not valid JSON, or contained an unknown field. |
+| `400` | `no_handler_found_exception` | No handler matches that route. |
+| `404` | `index_not_found_exception` | The index does not exist. |
+| `404` | `not_found` | The document does not exist. |
+| `404` | `resource_not_found_exception` | The document was not found when reading `_source`. |
+| `406` | `media_type_header_exception` | Unsupported `Content-Type` or `Accept`. |
```
jerguslejko, line 324: `media_type_header_exception` was missing, added in https://github.com/topk-io/ddb/pull/2041

topk-es/README.md
```diff
@@ -0,0 +1,360 @@
...
+
+Nothing here is silently ignored. Unsupported fields in a search body are rejected with
+`400 json_parse_exception`, which lists the fields that are accepted; unrouted endpoints
+return `400` with the bare `no handler found for uri` body:
+
+- Scroll and point-in-time search
+- `_update_by_query` and `_delete_by_query`
+- Highlighting and suggesters
+- `collapse` and `script_fields`
+- `date_histogram` and other bucket aggregations beyond `terms`
```
jerguslejko, line 349: will add docs for this in https://github.com/topk-io/topk/pull/565

jerguslejko (approved):

docs/sdk/topk-es/overview.mdx
davidmyriel: Good catch — fixed in 8e8130e. The whole page is on \`dims: 4\` now, and the quick-start documents got an \`embedding\` so the knn example actually has something to match. I also pointed the hybrid example at \`title\` instead of a \`body\` field no example index defines.

Pasted each block in order against a live cluster to confirm: create, bulk, bool search, knn, and hybrid all run clean.

topk-es/README.md
davidmyriel: Thanks — you are right, and I had this wrong rather than early: \`BoolQuery\` has no such field and \`deny_unknown_fields\` rejects it with a 400 (confirmed live). I filled it in from Elasticsearch convention instead of reading the struct.

Removed in 8e8130e, and added \`boost\`, which is accepted and was missing from the row. Worth re-adding to this table in #585 when the field lands.

topk-es/README.md
davidmyriel: Ah — so the docs were ahead of the server here. Thanks for adding it in topk-io/ddb#2041; the row stays accurate once that ships.

topk-es/README.md
davidmyriel: Sounds good — I will leave the \`date_histogram\` bullet as-is and let #565 remove it when the support lands, so the two do not conflict.

PR #585, opened 2026-09-01
[topk-es] bool minimum_should_match

jerguslejko:
add support for `minimum_should_match ` - supports int, string count, percent and negative forms

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ target/
 .vscode/
 .venv/
 __pycache__/
+topk-es/Cargo.lock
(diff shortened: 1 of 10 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ target/
 .vscode/
 .venv/
 __pycache__/
+topk-es/Cargo.lock
```
petertoth: intentional?

petertoth (approved):

.gitignore
jerguslejko: yep, we are ignoring topk-rs/js/py/es/sql locks. will generalize tho

PR #529, opened 2026-07-27
[cli] import

jerguslejko:
`topk import <source> [objects]` — discovers the source, prints the plan as a toml spec, confirms, creates the collections, upserts.

* databases — postgres, mysql, sqlite, mongodb, elasticsearch
* files — csv, tsv, json(l), parquet, arrow, avro, xlsx; local, s3/r2, gcs, azure, hugging face, http
* collections — `topk://<region>/<name>`, schema and indexes copied as-is
* spec — `--dry-run` prints the toml and sample documents, `-f spec.toml` re-runs an edited plan; connection details stay on the command line
* resume — `--resume <run>` skips finished collections and continues the in-flight one from its cursor

PR #589, opened 2026-09-03
[docs] document the import flags and object-store caveats

davidmyriel:
Follow-up to #529, testing its docs the way #583 tested the Elasticsearch ones. I built the CLI from `main` and ran imports against a public parquet dataset. The command works; the docs have three gaps.

### Five flags were never named

`--id`, `--partition`, `--limit`, `--concurrency` and `--batch-bytes` appear nowhere in the CLI docs, and `--spec` only ever shows up as `-f` inside one example. The section is written entirely as worked examples, so anything not in an example is undiscoverable. Adds a flags table.

`--id` is the one that bites: without it `--dry-run` writes `id = "<column>"` and refuses to preview rows, so the documented dry-run flow stalls on a flag the docs never mention.

### Object stores need credentials, even for public buckets

An unconfigured `s3://` fails with a duckdb error that does not point at the cause:

```
error: duckdb error: Invalid Configuration Error: Secret Validation Failure:
during `create` using the following: Credential Chain: 'config'
```

The same object over `https://` reads anonymously and works. The sources table read "credential chain; http anonymous", which suggested public buckets were fine — corrected, with the `https://` fallback shown.

### Globs don't work over `http(s)://`

duckdb rejects them outright. The sources table listed `s3://` and `http(s)://` together with no distinction.

### Also: `cli.mdx` was never re-synced

#529 committed `docs/cli.mdx` without running the sync, so two callouts still carry raw `> [!NOTE]` / `> [!WARNING]` markup and render as literal text on docs.topk.io. One is the warning about keeping credentials out of the spec. Running `make docs` fixes it; that's included here.

### Verification

Built `topk-cli` from `main` and exercised each behaviour against a live cluster: single file naming its collection, glob requiring `--to`, the `--dry-run` → `spec.toml` → `-f` round-trip, the run id in the header, `--filter` as SQL `WHERE`, `--partition` isolation, and a 500-row import checked for correct ids, field types and mapping. Every flag and default in the new table matches `--help`. Scratch collections deleted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/cli.mdx
+++ b/docs/cli.mdx
@@ -21,8 +21,11 @@ topk import ./books.parquet --region aws-us-east-1-elastica
 topk import postgres://user:pw@host/db 'public.*' --region aws-us-east-1-elastica
 ```
 
-> [!NOTE]
-> `topk import` discovers a schema, shows the plan, and imports on confirmation — see [import](#import).
+<Note>
+
+`topk import` discovers a schema, shows the plan, and imports on confirmation — see [import](#import).
+
+</Note>
 
 ## Authentication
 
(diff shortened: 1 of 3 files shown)


docs/cli.mdx
```diff
@@ -88,6 +91,16 @@ topk import gs://bucket/books.csv                             # az:// hf:// http
 
 A single file names its collection.
 
+An object store reads through its credential chain, and needs credentials even
+for a public bucket — an unconfigured `s3://` fails with a duckdb
+`Secret Validation Failure`. Anonymous reads go over `https://` instead:
```
jerguslejko, line 96: mhm, will look into this

PR #597, opened 2026-09-10
[topk-cli] login

petertoth:
* `topk login` with oauth using auth provider
* uses prod oauth config by default, override with env vars
* credentials are stored either via a file (`0600`) or via native os keyring
  * macos defaults to file due to unverified binary UX - default to keyring when we release signed binary
  * windows defaults to file due to file due to Credential Manager item size [limit](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Security/Credentials/constant.CRED_MAX_CREDENTIAL_BLOB_SIZE.html)
  * linux defaults to keyring
* `topk logout` clears persisted credentials if present


topk-cli/src/main.rs
```diff
@@ -87,30 +92,34 @@ async fn async_main() -> ExitCode {
...
-            match api_key {
-                Some(api_key) => {
-                    config::set_api_key(api_key)?;
-                    eprintln!("{} API key saved.", "✓".green());
-                }
-                None => {
-                    println!("Skipping authentication.");
-                }
+        Some(Commands::Login { no_browser }) => {
+            let auth = cli.endpoint.auth()?;
```
jerguslejko: lets keep the cmd body in `commands/login.rs`

topk-cli/src/main.rs
```diff
@@ -120,7 +129,7 @@ async fn run(cli: &Cli) -> anyhow::Result<ExitCode> {
         }
 
         Some(Commands::Logout) => {
-            config::clear()?;
```
jerguslejko: same here

topk-cli/src/auth/client.rs
```diff
@@ -0,0 +1,80 @@
+use std::time::Duration;
```
jerguslejko: https://docs.rs/oauth2/latest/oauth2/#example-asynchronous-api exists, let's not hand-roll this...

topk-cli/src/auth/login.rs
```diff
@@ -0,0 +1,92 @@
...
+            self.auth.store.lock().await?.save(session)?;
+            Ok(claims)
+        })
+        .await
+    }
+}
+
+#[cfg(test)]
+#[path = "../../tests/auth/login.rs"]
+mod tests;
```
jerguslejko: why declare test modules like this?

topk-cli/src/auth/store/file.rs
```diff
@@ -0,0 +1,77 @@
...
+pub(super) fn remove(path: &Path) -> Result<()> {
+    match std::fs::remove_file(path) {
+        Ok(()) => Ok(()),
+        Err(e) if e.kind() == ErrorKind::NotFound => Ok(()),
+        Err(e) => Err(e).with_context(|| format!("removing {}", path.display())),
+    }
+}
+
+/// Replace a file atomically so readers cannot observe partial data.
+pub(super) fn write_secret_file(path: &Path, content: &str) -> Result<()> {
```
jerguslejko: TIL this exists: https://crates.io/crates/atomic-write-file

topk-cli/src/auth/store.rs
```diff
@@ -0,0 +1,217 @@
...
+use crate::config::Config;
+
+mod file;
+
+const KEYRING_SERVICE: &str = "TopK CLI";
+const KEYRING_USERNAME: &str = "session";
+
+#[derive(Debug, Clone, Copy, Default, clap::ValueEnum, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum CredentialsStore {
```
jerguslejko: @petertoth let's start with file-based backed only, and add support for keyring in a separate PR?

jerguslejko (changes requested):

topk-cli/src/auth/login.rs
petertoth: it was there so that `tests/auth/....` tests can also test private interface. Moved those to their respective files but kept integration tests in `tests/auth/....` removing 
```
#[cfg(test)]
#[path = "../../tests/auth/login.rs"]
mod tests;
```

topk-cli/src/auth/store/file.rs
petertoth: I tried to replace our `write_secret_file` with `atomic-write-file` but the crate fails this test:
https://github.com/topk-io/topk/pull/597/changes#diff-43fc54151d10c12bbb64f810574c06e52c7415d5dd9544a88c11e9e6436d59d2R99-R113

topk-cli/Cargo.toml
```diff
@@ -88,3 +91,7 @@ uuid = { version = "1", features = ["v4"] }
 name = "import"
 path = "tests/import/main.rs"
 required-features = ["import"]
+
+[[test]]
+name = "auth"
+path = "tests/auth/main.rs"
```
jerguslejko: shouldn't be needed

topk-cli/src/auth/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    oauth_config: OAuthConfig,
+    client: BasicClient<EndpointSet, EndpointNotSet, EndpointNotSet, EndpointNotSet, EndpointSet>,
+    http: HttpClient,
+    store: SessionStore,
+}
+
+impl Auth {
+    pub fn new(config: &Config, config_dir: PathBuf) -> Result<Self> {
+        let oauth_config = config.oauth();
+        Ok(Self {
```
jerguslejko, line 30: the `oauth` client is spread out in multiple files now. We should wrap the primitives in our own client. something like:

```rs
struct OAuthClient { BasicClient, reqwest::Client, audience }

impl OAuthClient {
    fn authorize(&self, redirect_uri: &RedirectUrl) -> Authorization;
    fn exchange_code(&self, authorization: Authorization, code: String) -> Result<(Session, Option<AccessTokenClaims>)>;
    fn refresh(&self, refresh_token: String) -> Result<Option<Session>>;
}
```

and `Authorization` holds `Url`, `CsrfToken`, `PkceCodeVerifier` etc.


bunch of call-sites collapse, `Session::from_response` and `claims` move to `oauth`.

topk-cli/tests/auth/common/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub async fn request(&mut self) -> String {
+        timeout(Duration::from_secs(5), self.requests.recv())
+            .await
+            .unwrap()
+            .unwrap()
+    }
+}
+
+// Serialize logins sharing the three registered callback ports.
+pub static LOGIN: Mutex<()> = Mutex::const_new(());
```
jerguslejko: we should make the ports dynamic so tests don't have to serialize

topk-cli/src/auth/store.rs
```diff
@@ -0,0 +1,89 @@
+use std::fs::File;
```
jerguslejko: we should inline `topk-cli/src/file.rs` into this file.

topk-cli/src/auth/store.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    pub fn delete(&self) -> Result<()> {
+        file::remove(&self.store.credentials_file)?;
+        file::remove(&self.store.config_file)
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+struct Credentials {
```
jerguslejko, line 95: @petertoth WDYT about putting `store_key` in the filename (eg. `credentials/prod.toml`) instead of in the file?

topk-cli/src/auth/callback.rs
```diff
@@ -0,0 +1,190 @@
...
+        return StatusCode::BAD_REQUEST.into_response();
+    }
+    let code = match (params.code, params.error) {
+        (Some(code), None) if !code.is_empty() => Ok(code),
+        (None, Some(error)) => Err(anyhow!(
+            "{error}: {}",
+            params.error_description.unwrap_or_default()
+        )),
+        _ => return StatusCode::BAD_REQUEST.into_response(),
+    };
```
jerguslejko: nit: move onto `CallbackParams`

topk-cli/src/auth/callback.rs
```diff
@@ -0,0 +1,190 @@
...
+    let (response, received) = oneshot::channel();
```
jerguslejko, line 122: why do we need this callback channel? we could do just:

```rs
/// Serve the loopback callback until the provider returns a code, then shut down.
pub(super) async fn wait_for_code(
    listener: TcpListener,
    csrf: CsrfToken,
    timeout: Duration,
) -> Result<String>
```

petertoth: The `oneshot::channel()` sends the login result back to the browser so users can see something like:
`You're logged in. Close this tab` / `Login failed....`

jerguslejko: WDYT we say something generic, so there doesn't need to be a distinction? 

This surfaces _login_ failures in the browser, instead of "late" in the CLI. But there are also other ways login can fail (eg. store fails to save creds). So there's always a possibility of going back to the CLI and finding an error there.

topk-cli/src/auth/store.rs
petertoth: My initial idea was that we'd be storing only one session at the time for one oauth config (issuer). This would mean we have to maintain a session for each oauth config. It would allow `credentials/*` to grow unconditionally with changing any value in the oauth config.

jerguslejko: just per env no?

petertoth: Yeah per issuer which is settable by `TOPK_AUTH_DOMAIN`

jerguslejko: right, so we only have prod/stage/dev

petertoth: Yes, although it can be set to anything

topk-cli/src/auth/callback.rs
petertoth: I am fine with both. The current approach is better UX imo but it comes at extra cost in complexity.

topk-cli/tests/auth/common/mod.rs
petertoth: Made `--auth-callback-ports` a hidden flag with defaults for `login` command so tests can supply their own without needing to serialize.

topk-cli/src/auth/config.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub(super) fn oauth(&self) -> OAuthConfig {
+        OAuthConfig {
+            issuer: self.issuer.clone(),
+            client_id: self.client_id.clone(),
+            audience: self.audience.clone(),
+        }
+    }
+}
+
+fn parse_domain(domain: &str) -> Result<Url> {
```
jerguslejko: lets just parse as `Url` and try to use? this validation is a bit ad-hoc

topk-cli/src/auth/config.rs
```diff
@@ -0,0 +1,63 @@
...
+        default_value = AUTH_AUDIENCE,
+        hide = true,
+        global = true
+    )]
+    pub audience: String,
+}
+
+/// Resolved issuer, OAuth client, and API audience used for authentication.
+#[derive(Clone)]
+pub(super) struct OAuthConfig {
```
jerguslejko: this should move to `oauth.rs`

topk-cli/src/auth/login.rs
```diff
@@ -0,0 +1,59 @@
+use std::net::SocketAddr;
```
jerguslejko, line 1: WDYT about inline this in `commands/login.rs`? now that primitives are simpler, commands can hold the "control" logic

topk-cli/src/auth/login.rs
```diff
@@ -0,0 +1,59 @@
...
+impl<'a> Login<'a> {
+    pub(super) async fn new(auth: &'a Auth, ports: &[u16]) -> Result<Self> {
+        let addresses: Vec<_> = ports
+            .iter()
+            .map(|&port| SocketAddr::from(([127, 0, 0, 1], port)))
+            .collect();
+        let listener = TcpListener::bind(addresses.as_slice())
+            .await
+            .context("listening for the login callback")?;
+        let redirect_uri = RedirectUrl::new(format!("http://{}/callback", listener.local_addr()?))?;
```
jerguslejko: this could be a little helper

topk-cli/src/auth/store.rs
```diff
@@ -0,0 +1,288 @@
...
+    store: &'a SessionStore,
+    _lock: File,
+}
+
+impl LockedSessionStore<'_> {
+    /// Check readability before starting browser login.
+    pub fn prepare(&self) -> Result<()> {
+        read(&self.store.credentials_file)?;
+        Ok(())
+    }
```
jerguslejko: why do we need to check readability here? we don't do that when we read config for example

topk-cli/src/auth/oauth.rs
```diff
@@ -0,0 +1,136 @@
...
+        };
+        session_from_response(res, Some(refresh_token)).map(Some)
+    }
+}
+
+fn session_from_response(
+    res: BasicTokenResponse,
+    previous_refresh_token: Option<String>,
+) -> Result<Session> {
+    let expires_at = res
```
jerguslejko: nit: inline:

topk-cli/src/auth/oauth.rs
```diff
@@ -0,0 +1,136 @@
...
+            }
+            Err(e) => return Err(e).context("refreshing the access token"),
+        };
+        session_from_response(res, Some(refresh_token)).map(Some)
+    }
+}
+
+fn session_from_response(
+    res: BasicTokenResponse,
+    previous_refresh_token: Option<String>,
```
jerguslejko: nit: instead of optional arg, we can do `let mut session = ...` in `refresh`

topk-cli/src/auth/oauth/claims.rs
```diff
@@ -0,0 +1,71 @@
...
+            None => None,
+        }
+    }
+
+    pub fn account(&self) -> String {
+        match self.team() {
+            Some(org) => format!("{} in {org}", self.username()),
+            None => self.username(),
+        }
+    }
```
jerguslejko, line 72: nit: collapse all three? only consumer is `claims.account()` in `login`

topk-cli/src/auth/callback.rs
```diff
@@ -0,0 +1,199 @@
...
+        },
+    };
+    drop(shutdown);
+    // Flush the final response without allowing a slow browser to delay exit indefinitely.
+    let _ = timeout(SHUTDOWN_TIMEOUT, &mut server).await;
+    result
+}
+
+fn router(state: CsrfToken) -> (Router, mpsc::Receiver<Callback>) {
+    let (callbacks, receiver) = mpsc::channel(1);
```
jerguslejko: nit: take `callbacks` as arg so `router() -> Router`

topk-cli/src/auth/config.rs
```diff
@@ -0,0 +1,63 @@
...
+
+const AUTH_DOMAIN: &str = "topk-prod.us.auth0.com";
+const AUTH_CLIENT_ID: &str = "2LqddiN2N5fQplfMP2MIYPHM6ttFNeaG";
+const AUTH_AUDIENCE: &str = "https://api.topk.io";
+
+#[derive(clap::Args, Clone)]
+pub struct Config {
+    #[arg(long = "auth-domain", env = "TOPK_AUTH_DOMAIN", default_value = AUTH_DOMAIN, value_parser = parse_domain, hide = true, global = true)]
+    pub issuer: Url,
+    #[arg(long = "auth-client-id", env = "TOPK_AUTH_CLIENT_ID", default_value = AUTH_CLIENT_ID, value_parser = NonEmptyStringValueParser::new(), hide = true, global = true)]
```
jerguslejko: fmting?

topk-cli/tests/auth/session.rs
```diff
@@ -0,0 +1,536 @@
...
+    let auth = Arc::new(server.auth(dir.path()));
+    server.reply(200, response(0, Some("refresh"))).await;
+    seed(&auth).await;
+    server.request().await;
+    let refresh_auth = auth.clone();
+    let refresh = tokio::spawn(async move { refresh_auth.access_token().await });
+    server.request().await;
+    let logout_auth = auth.clone();
+    let mut logout = tokio::spawn(async move { logout_auth.logout().await });
+    assert!(timeout(Duration::from_millis(75), &mut logout)
```
jerguslejko, line 56: we should use `test_context`...

topk-cli/src/auth/oauth/claims.rs
petertoth: I split it for upcoming `whoami` command which displays these separately

topk-cli/src/auth/store.rs
petertoth: it was primarily used for keyring store but removing it for now

topk-cli/src/auth/login.rs
petertoth: I tried moving it into the command but that meant exposing the oauth client and session store, or adding methods on Auth to wrap them. I would prefer having the callback, token exchange, and store together in auth module.

topk-cli/src/auth/callback.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+async fn callback(
+    State(state): State<CallbackState>,
+    Query(params): Query<CallbackParams>,
+) -> Response {
+    let Some(code) = params.validate(&state.state) else {
+        return StatusCode::BAD_REQUEST.into_response();
+    };
+    let (response, received) = oneshot::channel();
```
jerguslejko: these are usually named `rx/tx`

jerguslejko (approved):

PR #602, opened 2026-09-18
[es][sql] read-your-writes

jerguslejko:


topk-sql/src/stmt/mod.rs
```diff
@@ -146,6 +144,15 @@ impl TryFrom<SqlStatement> for Statement {
...
 }
 
+// The only `RETURNING` TopK supports on a write: the LSN the write landed at.
+pub fn returning_lsn(returning: Option<Vec<SelectItem>>) -> Result<bool, Error> {
+    match returning.as_deref() {
+        None => Ok(false),
+        Some([item]) if item.column_name() == "_lsn" => Ok(true),
+        Some(_) => sql_unsupported!("RETURNING other than `RETURNING _lsn`"),
+    }
+}
```
petertoth, line 156: should this handle a case where syntax such as `RETURNING 123 as _lsn` is explicitly rejected ?

topk-sql/src/stmt/delete.rs
petertoth, line 39: does this also handle delete partition with lsn ?

jerguslejko: we now return `sql_unsupported!(returning_lsn, "RETURNING _lsn on a partition delete")`

topk-sql/src/stmt/mod.rs
jerguslejko: good catch, we now reject aliases in `RETURNING` position

petertoth (approved):

